PR #3, opened 2016-01-04
Adjustments after move to organization

stefandollase:
This pull request contains fixes for the repository, so it works correctly after the move to toolbox4minecraft/amdist. It also contains changes in the documentation.


JMoVS:
very great work about the new org and the changes. Congrats!

PR #41, opened 2016-01-24
Make End cities more accurate

Treer:
Fixes bug where the influence of all nearby islands wasn't taken into account.


stefandollase:
I was wondering about this when I ported the code. I must have missed the opportunity to add a comment that is asking whether it is correct. Good that you checked the ported code :-)

Just another note: I would prefer you to commit changes on separate branches and then create a pull-request from the branch, as described in the [Github Flow](https://guides.github.com/introduction/flow/). This also enables you to work directly in the `toolbox4minecraft/amidst` repository: Since you are a collaborator you can push directly to this repository, so you don't need to fork it.

Nevertheless, this pull-request looks good, so I will merge it.

Treer:
Thanks for the note - I assumed github forks were a fancy wrapper around git branches, I assumed wrong.

PR #42, opened 2016-01-25
Test to ensure that the behavior of the world object does not change

stefandollase:
To prevent the need to make the minecraft jar file and libraries available on travis ci, this test uses previously generated testdata to ensure that the generated world icons are still the same as when the test data where generated. It is true, that this uses the production code to generate the test data, however, this is not an issue, because the test data are stored and not generated every time the test is executed. All in all, the test checks whether the generated world icons have changed since the test data where generated. It does not ensure, that the world icons are correct (meaning like minecraft generates them).

The test:
* loads the biome data from the test data
* loads the world icons
* creates a fake minecraft interface that provides the stored biome data
* creates a world object using the fake minecraft interface
* generates all world icons using the world object
* compares the generated world icons to the loaded world icons

The test data contains all biome data in the test area as well as all world icons. The test area is 10 fragments in all directions from 0, 0 ([-5120, -5120] to [5120, 5120]). The world icons are generated for the inner 10 - 1 = 9 fragments ([-4608, -4608] to [4608, 4608]). This prevents the world icon producers to request biome data that have not been stored. The test data are stored in a zip-file to prevent them from taking up to much space. One world requires about 6.5 MB (compressed). Thus, we should not add to much test worlds and not change them very often to prevent the repository from getting too big. An alternative would be stored the testdata somewhere else than in the repository. However, they have to be available for travis ci, so this will be the solution for now.

* The End Cities are currently not included.
* The current seed is "amidst-test-seed" using the default world type and minecraft version 1.8.9.
* The test data are generated by running the test in the class DevToolRunner.


stefandollase:
@Treer and @BigAlanM maybe one of you knows a seed that is good for testing the world icon generation? As described above, this test uses a limited area around 0, 0 for a given Minecraft version and a given seed. So, if you know a combination of seed and Minecraft version that generates a sufficient amount of world icons of each type in the give region, that would be really helpful. It would be even more helpful if the seed includes world icons that are easily generated wrong, for example because they are on a border between two biomes.

Treer:
Sounds like a job for a seedfinder search, which BigAlanM knows more about.

FWIW -1364077613 was interesting because it misses an ocean monument if the quarter-resolution map is used for the structure test, but there's no nearby igloos. Often I use 1 (it's easy to type), which by luck happened to illustrate the [minecraft bug in how the new 1.9 strongholds are placed](http://i.imgur.com/x9OrkhL.png) (awkwardly reported [here](https://bugs.mojang.com/browse/MC-92289)). But yeah, a seed wth all the biomes makes more sense.

stefandollase:
I got the test world size to under 50 kB. Thus, we no longer need to be super economically with the test worlds.

stefandollase:
@Treer Thanks for the seed suggestions. I added all of them, since one test world now became relatively cheap. After I added all the data from the world object it is still less that 100 kB per world. However, we should nevertheless keep an eye on the number of test worlds to keep the tests and thus the automatic build process quick.

PR #44, opened 2016-01-26
Add graphics acceleration indicator to Debug widget

Treer:
Might be a bit of an overkill, since I *suspect* graphics are accelerated for nearly everyone, but I'd like to keep an eye on it during development.

There's now a dot on the FPS widget that's green if everything's accelerated, and blue if it's not. The Debug widget lists the value as a percentage.


Treer:
The [network graph](https://github.com/toolbox4minecraft/amidst/network) suggests I'm still not doing this branch thing properly. I think it's before I forgot to pull to my computer before creating the next branch. Let me know if it's messed up from your end.

stefandollase:
I will look into this now. I think it is a good idea to add it to the debug widget. The numbers in there are generally confusing, so one more won't hurt :-P

Regarding the branching: You already figured it out. Before you create a branch you need to pull and checkout master. I will try to fix it for the messed up branches by rebaseing and force pushing. This might cause problems on your end, however these problems should vanish as soon as the pull request is merged.

stefandollase:
I removed the dot from the FPS widget. This will hopefully reduce the confusion. Other than that I do like the idea to include it as a debug information. This should help us to sort out the issue with the unaccelerated graphics if it is an issue for many users.

PR #45, opened 2016-01-26
Display Void texture in parallax.

Treer:
Changes the void so it's no longer scaled and translated with the End islands, instead they "float" in a parallax layer above it.

You might have made this change already, or have better ideas how to do it.


stefandollase:
I removed the old code, since it was unused. Other than than, I would have implemented it the exact same way :-)

stefandollase:
This pull request wants to merge to another branch than master, so I merged it from the client side into master.

PR #46, opened 2016-01-29
Add support for snapshots when -mcpath is a network path

Treer:


stefandollase:
I did not yet get the exact purpose of this. Is the network path a windows thing? It looks like it because of the `\\` in the path. Also, I did not yet notice a special `.minecraft_snapshots` directory, which might be related to the fact that I create separate profile directories for all my profiles, anyway. Do you consider this pull request a bug fix of a feature? Maybe that helps me to sort things out in my head.

Treer:
I may have messed up here, I thought the .minecraft_snapshots directory was a new thing created by the Mojang launcher. Amidst4 wasn't finding my recent snapshots, because they were going to that directory - which I didn't specify.

But I've just tested creating snapshot profiles on another computer, and it placed them in the .minecraft directory, plus googling minecraft_snapshots hasn't turned up much, so I'm starting to suspect perhaps .minecraft_snapshots was created by me years ago and forgotten about, and when I created profiles for recent snapshots it duplicated from an old snapshot profile from years ago.

I'll keep looking into it, but I would have expected it to show up on Google if it was a thing. Will check creation dates when I get home.

UNC paths apparently exist in posix (starting with // instead of \\) but I think Windows is the only place where they are commonly used.

Treer:
AmidstExporter (and presumably old Amidst?) doesn't have this issue because it's not smart enough to check the gameDir path from the profile, but probably has a whole bunch of other bugs caused by that instead. 

So I think this is probably unfixable, but will affect few people as most users of -mcpath will still be running Amidst on the same computer they play games on.

Treer:
OK, I messed up. My `.minecraft_snapshots` folder was created 2 years ago, so I must have done it. This pull request doesn't solve anything.

stefandollase:
Okay, stuff like this happens. So can I completely remove the changes introduced by this pull request?

Treer:
yup, kill it.

stefandollase:
Done with #52

Issue #59, opened 2016-02-03
[MC-92289] Extra strongholds spawn in positions Mojang didn't intend


stefandollase:
This issue provides a place for further discussion about how we handle [MC-92289](https://bugs.mojang.com/browse/MC-92289). We already talked about it in #42 and #33.

Quoting @Treer's response to my question from #33:

> > Do you think we should do anything before we release the stable version?

> I think wait until we know what's happening with the stronghold locations bug in the 1.9 snapshots. If it looks like 1.9 is going to contain this bug then release Amidst as it is, if it looks like they will fix it before 1.9 then hold off the release until we find out what they're changing it to. Either way we should be able to release a stable version before 1.9 comes out.

Treer:
Searge just marked this as fixed, so when they release the next snapshot we'll be able to update the Stronghold algorithm.

Treer:
I've committed the updates, I've somehow committed straight into master instead of it creating the Issue_MC-92289 branch here, and I think any attempt by me to clean this up is going to make things worse, but the code is there - feel free to move it into a seperate branch.

The Amidst4 refactoring has made this change quite hard, related code has been broken up and hidden away in various methods and patterns, I've tried to emulate Amidst4's existing structure with this change and it means I don't envy anyone trying to decypher how this stronghold code works. This could mean I haven't done it the way you would and you might like to refactor it yourself. For perspective, here is the same change in AmidstExporter: https://github.com/Treer/AmidstExporter/commit/a9d9fd4ca22322fec9e1878175dd2f3bee56bf1e

PR #60, opened 2016-02-13
Implement new stronghold position algorithm

Treer:
This is the update for #59

I somehow committed straight into master initially instead of it creating the branch, I think I've cleaned it up but I'm leaving this as a note incase it caused any weirdness.

The Amidst4 refactoring has made this change quite hard - related code is broken up and hidden away in various methods and patterns, I've tried to emulate Amidst4's existing structure with this change and it means I don't envy anyone trying to decypher how this stronghold code works. Feel free to refactor it. For reference, and perspective, here is the same change in AmidstExporter: https://github.com/Treer/AmidstExporter/commit/a9d9fd4ca22322fec9e1878175dd2f3bee56bf1e


stefandollase:
Thanks for figuring it out :-) Also, you got the branching right.

I see what you mean when you say that it is hard to make an algorithm like this modular. I also think that it is not easy to understand, which is really bugs me. However, I think this is one of the points that Amidst cannot easily get around. This is due to the fact that it supports all Minecraft versions. The algorithms change every now and them. While each Minecraft version always contains only one version of each algorithm, Amidst has to keep all versions of all algorithms, forever. This means, that we get a huge amount of code duplication or we have to make the algorithms modular like this, which leads to code that is hard to read. Both is not a good thing and we should make the decision for each individual piece of code.

When I refactored the code, I noticed that there was lots of code duplication. It was very hard to tell what exactly is duplicated and what is different. So, I removed lots of duplication. Now, the differences are really easy to see, however the algorithm itself is harder to read. I still think this was a good change. Since for Amidst it is important to easily see the differences between different versions. Also, most of the time you do not need to understand the algorithm as a whole.

Another point is that if there is an bug in the duplicated code, then you have to fix it everywhere and not only once, which is quite annoying. I am also talking about the general structure generation algorithm here, which was duplicated a lot in the original Amidst.

As I said, we should make the decision from case to case.

I think that is what you meant, did you? If there is another part of the architecture that is too complicated, we can always talk about it. Also, as I said, you can always ask me, if you do not understand parts of the code. I will be happy to help.

Finally, lets have a closer look at what I think was the problem here: You had to use the version features to make the decision about the stronghold algorithm, so you created separate classes for the buggy and the fixed algorithm and created a base class which contains the actual algorithm. An alternative would be to pass a flag to the constructor of the stronghold producer, which tells it to use the buggy or the fixed algorithm and maybe a second flag to tell it whether to use the original 3-stronghold algorithm of the new 128 algorithm. Then you would have all required information available in the one stronghold producer class. You would just pass different flags when it is initialized in the class DefaultVersionFeatures. For an example of this, have a look at the initialization of the mineshaft algorithm. The resulting code would be quite similar to the change you made in AmidstExporter. However, there would still be some differences:

* The flags are only set once, when the world is created instead of every time the stronghold creation method is called.
* All information about what is changed by the recognised version is concentrated in one place, the class DefaultVersionFeatures.
* The old Amidst and AmidstExporter have one big method for the complete algorithm, which is generally considered bad practice. Thus, I split the algorithm in smaller parts.

All in all, I do not know which code would be better to read and maintain. The one with multiple classes and inheritance or the one with the flags. I would have to try it by myself. However, I think the current code is good enough for now. Thanks again :-)

stefandollase:
Now, lets talk about the changes I made.

There where some multiple enum values for the same magic string in the enum RecognisedVersion. One duplication was introduced by your changes in this pull request, but there where more duplications that where already contained in the old Amidst. I removed the duplications and added a test to ensure there won't be duplications in the future.

Also, Mojang seems to have recently added some 1.7 versions back to the launcher. Sadly, they all map to the same magic string as 1.7.9, so I only added them to the comment.

Then, I removed the introduced version from the stronghold producers and placed the version and documentation directly into the class DefaultVersionFeatures, since that is the place to store all of the information related to the RecognisedVersion. That was the reason to create the class.

Besides minor formatting and refactoring changes, I adjusted the test data to the new stronghold algorithm. This was what took the most time, since I stumbled across the issue [MC-95612](https://bugs.mojang.com/browse/MC-95612). Sadly, that was exactly the snapshot I used originally to create all the test data. The fact that there was no new magic string for the fixed algorithm did not help either.

Before I noticed that the changes of most of the test data was related to the issue linked above, I had another suspicion. I thought that the changed stronghold algorithm would request and generate the test data in a different order than the old algorithm. This was still a bug with the test data storage. It stored the test data pretty much in a random order. Thus, the test data file might change even tough it contains the same data in a different order. Now, I defined a fixed order for the test data storage, so this is now also fixed.

stefandollase:
Since this is now all working and tested, I merge it into master.

Treer:
TriFunction does make it easier to read than doing a FactoryFactory - I was looking into making a TriFunction, but then read about why Java doesn't provide one and it put me off, but this is definitely the right solution for this situation. 

To be clear, I don't have a problem with a subclass for each version implementation of an algorithm - I started doing that myself using inner classes before I noticed that you'd already created a factory structure for the same situation with MineshaftAlgorithm. 

The decisions that seemed odd to me were
 * Not keeping algorithm information in one place, e.g. StrongholdProducers don't know which Minecraft versions they were written for, or what biomes they apply to. I can see that you're instead trying to keep all *version history* information in one spot rather than keep the algorithm information together, and seperating stronghold biomes from the algorithm was an unfortunate side-effect of that. So I see now this is 6 of one and hald a dozen of the other - you don't need to defend this decision.

* Breaking code into lots of small methods, with the difficulty of following that code compounded by the methods potentially being polymorphic so not necessarily clear which is being invoked at runtime. It seems to be a general style, but as an example, instead of each StrongholdProducer class implementing a complete stronghold algorithm, we have a base implementation with lines like `angle += getAngleDelta()`. It saves code duplication, but in this situation code should be duplicated - a bug in one might not be a bug in the others, similarities between the different versions of the algorithm are luck and we needn't preserve the "pattern" of the agorithm, which could change completely next time. 

I like the new StrongholdProducer class names.

I wasn't sure how deliberate the default eclipse formatting was :D I'm not a fan of eclipse's default formatting - nobody uses an 80 column screen any more, and a few of its  "wrap before operator" settings are bad form, but I'm not really a Java coder and going with the default formatting style associated with a language is usually the best way to avoid everyone having different style preferences.

stefandollase:
Yes, having to add an extra `TriFunction` interface is nothing I like to do, but for this case it seemed to help the readability.

I do agree with you that it is really hard to follow an algorithm if parts of it are overwritten by subclasses. And yes, this might indeed be the right place for code duplication. However, I think in general it is easier to read the code if it is split into smaller parts. The only thing that makes it hard is the inheritance. The good thing about splitting it up is that you have a name for the extracted part of the code. So, it is not only easier to understand the extracted part of the code, but it is also self-documenting the compound code. Instead of adding little comments all over the place you just extract a method that has a meaningful name. But it might be correct that I overuse this technique from time to time.

I can also imagine a better code formatter. However, this was the most accessible. Everything else would involve additional tools that need to be set up by all developers. I did not want to do this, so I went with the default eclipse formatting. It is better to have some formatting than to have none. Sadly, there is no real standard Java formatting, but many different formatters that are all widely used.

May I ask you about what is you main programming language? My hunch is that it is C#, because of the first uppercase letter for method names, that you tend to use from time to time :-P

Treer:
Yeah, currently C#, and coincidently some Delphi a few years before that, and the same guy designed both languages so it's been a long time since I was using a language that didn't capitalise method names. Hmm, now that other people will be reading this Java code I should read a java style guide.

stefandollase:
Most of the time your coding style is actually pretty near to how Java code should look like. Here is what I noticed so far. Java uses:

* `lowerCaseMethodNames()`
* `ALL_UPPER_CASE_WITH_UNDERSCORE_CONSTANT_NAMES` (everything declared as `static final`)
* mostly no underscores in all other identifiers (this might just be a personal preference of mine)
* Javadoc comments for classes, fields and methods instead of comments in the methods (this might also be a personal preference)
* always `{}` for `if`, `else`, `while` and `for`
* no single-line methods (without line-breaks)

Javadoc comments are multiple comments with two asterisks after the opening slash:

```
/**
 * I am a Javadoc comment.
 */
```

```
/*
 * I am a normal multi-line comment.
 */
```

```
// I am a single-line comment.
```

I tend to use some more formatting rules, but I do not want to bore you :-P

Issue #81, opened 2016-02-18
User reports "white screen" with v4.0


BigAlanM:
From: http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-tools/1262200-v3-7-amidst-strongholds-village-biome-etc-finder?page=254#c5119
****************************
Hrmm, I'm stumped! The "new" version of Amidst - the beta 8 version of 4.0 won't work for me at all. I'm on PC, and have tried with several different versions of minecraft - 1.8.8 1.8.9 1.0 pre1 and some snapshots from 2014. I get the profile selection screen, hit whichever profile I want to test, and then it just opens a white screen with no menus or anything, just the white window. No error messages or anything, just a white screen showing me how dirty my monitor is =P

Tried with what I think is the "old" amidst (amidst exporter) and that one opens up normally and can be used.

Anyone have any ideas what's going wrong?

*edits* I did try both the .exe and the .jar just in case it made a difference somehow!
****************************
I don't have anything constructive to tell him.
Old version of JAVA?

Treer:
Yup, I have the same issue on my PC at work: AmidstExporter works, Amidst4 is just a white window.

Googling around it sounded like an old version of Java, but my Java was up to date when I first noticed it, and I've just updated Java now to double check - no change.

My money is on this being a video acceleration issue. My work PC has a cheap video card and whatever drivers Windows 10 saw fit to use. 

I never figured out [why AmidstExporter was accelerated on a virtual machine at home, but not Amidst4](https://github.com/toolbox4minecraft/amidst/issues/43#issuecomment-174761525), but these issues are probably related.

Treer:
OK, definitely looking like a driver bug - if I drag the window across to the second monitor as soon as it appears, the content of the window appears.

BigAlanM:
The user updated his video driver and that didn't help.

http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-tools/1262200-v3-7-amidst-strongholds-village-biome-etc-finder?page=254#c5121

Talyda:
Hiya!  I'm the one having the problem (and it's she, not he BTW - but I'm used to the assumption after being a gamer for many many years =P)  If there's any other info I can give to help work it out, just yell.  Been using Amidst for a long time, but the "old" one (Skidoodle's), not this new one. I may actually go use my husband's PC while he's at work today just so I can see what this one is like - and find the right seed for what I am wanting to do! =)

[...]

Issue #43, opened 2016-01-25
Low framerate and high CPU usage


Treer:
If I run the original amidst and make it full screen, it sits at 50fps using little CPU.

If I run Amidst4 and make it full screen, it sits at about 25fps using full CPU.

(I believe Amidst is limited to 50fps, and obviously the Amidst4 framerate will depend on the computer you run it on - mine is a virtual machine running on an old underpowered computer, which makes the difference noticeable. FWIW the VM supports hardware graphics acceleration)

Treer:
I did some profiling but am not yet familiar enough with the code to figure out why ImageDrawer.draw() is creating such a load.

I was thinking perhaps the graphics routines aren't being accelerated for some reason, so it's all being done with CPU. As far as I can tell, Amidst4 *is* enabling graphics acceleration, but [point 2 of this post](http://stackoverflow.com/a/13832805/3343347) indicates that the way BufferedImages are used can affect whether they are accellerated. This whole theory could be a red-herring, but it's as far as I got on the problem tonight.

stefandollase:
Amidst is indeed limited to 50 fps. It is odd, that we have a performance issue, especially because I do not know about any specific change that I made, that could cause this issue. I did some testing with the accelerated buffered images. On my machine, most of the images are being accelerated. Here is the code I used to test this:


	private volatile int accelerated = 0;
	private volatile int counter = 0;
	private final GraphicsConfiguration gc = GraphicsEnvironment
			.getLocalGraphicsEnvironment().getDefaultScreenDevice()
			.getDefaultConfiguration();

	private void log(BufferedImage image) {
		if (image.getCapabilities(gc).isAccelerated()) {
			accelerated++;
		}
		counter++;
		if (counter % 10000 == 0) {
			System.out.println("accelerated: "
					+ (100.0d * accelerated / counter) + "%");
		}
	}

Place this in the class `ImageDrawer`. Then replace the line:

		g2d.drawImage(fragment.getImage(declaration.getLayerId()), 0, 0, null);

with

		BufferedImage image = fragment.getImage(declaration.getLayerId());
		log(image);
		g2d.drawImage(image, 0, 0, null);

I get about 99% accelerated images. Maybe, you can use this to find out whether the acceleration is the problem.

[...]

PR #89, opened 2016-02-20
added setting graphicsAcceleration

stefandollase:
This is a workaround for the white window issue described in #81 and #43. It adds a setting to completely disable the Graphics Acceleration. You can activate this setting by using the key stroke **CTRL + A** in the main window. You will not get any feedback after doing do. Also, you need to restart Amidst for the changes to take effect.

I am not sure whether we should publish this. It might lead to issues like the following:

The user disables Graphics Acceleration without knowing it by accidentally clicking around in the settings or by hitting **CTRL + A**. After the next restart, the Graphics performance is very bad. He does not know were this comes from and is lost. He might report a bug, but he might as well ignore it and live with the bad performance. This would be a great reduction in usability.

### DIFF
--- a/src/main/java/amidst/Amidst.java
+++ b/src/main/java/amidst/Amidst.java
@@ -3,6 +3,7 @@ package amidst;
 import java.io.File;
 import java.sql.Timestamp;
 import java.util.Date;
+import java.util.prefs.Preferences;
 
 import javax.swing.JOptionPane;
 import javax.swing.SwingUtilities;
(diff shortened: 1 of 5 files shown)


stefandollase:
I added logging and a message dialog to this. The message dialog informs the user when he disabled the setting. It also warns about the drop in performance and tells him how to re-enable it. There is no message dialog when the setting is turned to enabled. This is to clearly get a feedback if you cannot read the text, because everything is white.

stefandollase:
Note, that this only disables the explicit activation of the graphics acceleration. It seems like disabling this has no effect on the graphics acceleration taking place on my Ubuntu machine.

stefandollase:
I am now a lot less concerned about the usability issue of accidentally disabling the Graphics Acceleration. Thus, I will integrate this, since it should allow users with the white window issue to use Amidst. The usage will not be fun, since it is very slow, but the usage will at least be possible.

stefandollase:
This got removed.

PR #94, opened 2016-02-20
Fix 2D acceleration issues on Windows

### LINKED ISSUE #43: Low framerate and high CPU usage

### LINKED ISSUE #81: User reports "white screen" with v4.0

Treer:
Fixes #43, and probably fixes #81

### DIFF
--- a/src/main/java/amidst/Amidst.java
+++ b/src/main/java/amidst/Amidst.java
@@ -125,16 +125,29 @@ public class Amidst {
 		return new AmidstSettings(Preferences.userNodeForPackage(Amidst.class));
 	}
 
+	/**
+	 * WARNING: This method MUST be invoked before setLookAndFeel().
+	 * The sun.java2d.* properties have no effect after setLookAndFeel() has 
+	 * been called.
+	 */
 	private static void enableGraphicsAccelerationIfNecessary(
 			Setting<Boolean> graphicsAcceleration) {
 		if (graphicsAcceleration.get()) {
 			Log.i("Graphics Acceleration: enabled");
-			setAndLogProperty("sun.java2d.opengl", "True");
+			if (!isWindows()) {
+				// We don't set OpenGL on Windows because Direct3D is better   
+				// supported, and the default.
+				setAndLogProperty("sun.java2d.opengl", "True");
+			}
 			setAndLogProperty("sun.java2d.accthreshold", "0");
 		} else {
 			Log.i("Graphics Acceleration: disabled");
 		}
 	}
+	
+	private static boolean isWindows() {
+		return System.getProperty("os.name").toLowerCase().contains("win");
+	}	
 
 	private static void setAndLogProperty(String key, String value) {
 		System.setProperty(key, value);
--- a/src/main/java/amidst/Amidst.java
+++ b/src/main/java/amidst/Amidst.java
@@ -20,7 +20,6 @@ import amidst.gui.crash.CrashWindow;
 import amidst.logging.FileLogger;
 import amidst.logging.Log;
 import amidst.mojangapi.file.DotMinecraftDirectoryNotFoundException;
-import amidst.settings.Setting;
 
 @NotThreadSafe
 public class Amidst {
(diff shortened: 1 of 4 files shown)


stefandollase:
I will put this here, just for future reference when we all forgot about this:

| Version                 | LAF guard   | is LAF never set (buggy guard)? | what comes first? |
|-------------------------|-------------|:-------------------------------------:|:-----------------:|
| Amidst v3.7             | !isOSX()    |                   no                  |        LAF        |
| AmidstExporter          | !isOSX()    |                   no                  |        LAF        |
| up to Amidst v4.0-beta8 | isWindows() |                  yes                  |        LAF        |
| since Amidst v4.0-beta9 | isWindows() |                   no                  |       Java2D      |

The switch from `!isOSX()` to `isWindows()` was meant to only switch Linux over to the default Java LAF. However, `isWindows()` contained a bug which caused it to always return false.

Combine this with the requirement that the Java2D variables, to be effective, need to be set before the LAF. Then this becomes a really hard bug to track down.

The beta9 fixed the ordering and the `isWindows()` guard. However, regarding the Java2D properties, this did not cause a change in behavior on windows:

* before, the LAF was never set and thus the Java2D properties were always effective
* after, the Java2D properties were set first and thus were always effective

The solution to all the issues now seems to be to not enable OpenGL on Windows, using the Java2D variables. This is because Windows uses Direct3D by default, which is better supported than OpenGL. Activating OpenGL when Direct3D is already active seems to cause bugs.

Of cause, there are windows machines without Direct3D, such as mine. They could profit from the activation of OpenGL. However, activating OpenGL on my windows machine causes the complete  Java GUI to behave very strange and buggy, as described [here](https://github.com/toolbox4minecraft/amidst/issues/43#issuecomment-186041640) and [here](https://github.com/toolbox4minecraft/amidst/issues/43#issuecomment-186045074). This behavior is not limited to Amidst. I can reproduce it with a simple JFrame containing any Component: As soon as I activate OpenGL, it gets buggy.  This does only happen on Windows, activating OpenGL on Linux works fine for me.

Thus, this pull-requests will:

- [ ] the LAF already is set to the default Java LAF on all Operating Systems
- [ ] disable OpenGL support on all windows machines
- [ ] remove the setting to disable the Graphics Acceleration: I never was happy about putting it in
- [ ] remove Java2D property logging: it does not reveal any information

---

Note, that this was all about Windows. However, I can say that on my Ubuntu machine, the debug widget tells me that the Graphics are accelerated, even if I do not activate OpenGL using the Java2D property. I do not know what is causing this. Explicitly setting the variable to false also does not disable the acceleration (is false even a valid value?). I was under the impression that Direct3D was Windows only technology, so I guess it is not Direct3D which is causing the acceleration.

For now, I will let OpenGL enabled for Linux and OS X.

---

One question remains: How did the old Amidst not have these problems? I guess, the answer is, that the old Amidst never had OpenGL activated on Linux and Windows, but only on OS X, since OS X was the only Operating System which did not set a custom LAF before the Java2D properties were set. This is also a good reason to disable OpenGL on Windows: Amidst v3.7 never seemed to have these problems.

If we receive similar bug reports from Linux users, we should probably just go ahead and disable OpenGL on Linux, too, since this is the old Amidst v3.7 behavior. We should also consider to disable it for all platforms, if we keep getting further bug reports.

stefandollase:
I also disabled OpenGL on Linux, see #104.

Issue #98, opened 2016-02-21
Menu shortcuts use 'ctrl' on OS X


moulins:
Because the menu shortcut keymask is hardcoded inside AmidstMenuBuilder.java and LayersMenu.java, all the shortcuts use ctrl instead of cmd on OS X.

I already have a fix commited on my fork which should be ready to merge. (https://github.com/moulins/amidst/tree/menu-fix-shortcuts)

stefandollase:
Thanks for contributing code :-) Did you read the wiki entry [Writing Code](https://github.com/toolbox4minecraft/amidst/wiki/Supporting-the-Development#writing-code)? When you did, please [create a pull-request from your branch](https://help.github.com/articles/creating-a-pull-request/).

moulins:
Okay, I made a pull request with my changes. :) ( #99 )

PR #99, opened 2016-02-21
Fix menu shortcuts on OS X

moulins:
This pull request fixes the issue #98.


stefandollase:
Thanks, this should improve the user experience for OS X users :-)

Issue #113, opened 2016-03-03
Ocean monuments were updated in 15w46a


Treer:
[from reddit](https://www.reddit.com/r/Minecraft/comments/48e3he/amidst_for_19_now_available_locate_igloos/d0kngpk):

> Looks like you missed the change to Ocean Monument generation in 15w46a: instead of a single deep ocean biome at the very center, it now requires deep ocean everywhere in a square with inradius 10.
>
> For example, in seed -3189899870270394863 the monument Amidst shows at 808, 1224 doesn't generate anymore.

I haven't investigated this yet, but it sounds likely.

PR #118, opened 2016-03-05
Update OceanMonumentLocationChecker to reflect changes in 15w46a

Treer:
This is the fix for #113

Bug [MC-65214]( https://bugs.mojang.com/browse/MC-65214) was fixed in Minecraft 15w46a, and that fix changed where Ocean Monuments can appear.

Examples: In seed -3189899870270394863, the monuments at (808, 1224) and (-856, 184) are eliminated by this fix.

I've confirmed the example monuments using Minecraft 1.8.9 vs 1.9, but didn't update the CI tests - I haven't looked into how they work.


stefandollase:
Are you familiar with executing the devtools? To fix the failed checks, you basically just execute the devtool [generateWorldTestData](https://github.com/toolbox4minecraft/amidst/blob/ocean_monument_fix/src/test/java/amidst/devtools/DevToolRunner.java#L40). After you did this, you need to update your eclipse project, because it might not automatically notice that the test data zip files were changed. You do this by selecting the project itself and then pressing F5. Now you can execute the tests again and they should all pass.

Make sure only the test data zip files, which should be affected by the change are actually changed. You should verify the Minecraft version id and the type of the test data (stronghold, village, ...).

If you run into any issues with this, please ask again.

Treer:
How do you execute that generateWorldTestData?

stefandollase:
Ah, I thought I already saw you use the devtools, but I guess not. Each devtool is now a JUnit test method in the class `DevToolRunner`. To execute a JUnit test class, you select the class in the Eclipse Package Explorer and

1. right click the selected item
2. Run as
3. JUnit Test

However, you will notice that nothing will be executed when you do this for the `DevToolRunner`. This is because these tests should not be executed by Travis CI. Most of them would not even work there. They are supposed to only run on the developer machines. To actually execute one of the devtools, you have to remove its `@Ignore` annotation and then execute the test as described above. Remember to add the annotation back after you are done.

To execute the actual tests, you can either execute `mvn test` (via the command line or the Eclipse menu) or you can select the source folder `src/test/java` in the Eclipse Package Explorer and then follow the three steps above. This finds and executes all JUnit tests in this source folder. Travis CI uses the `mvn test` command.

Treer:
Thanks. Now that I have it working it's quite a nice setup.

How come the biome data changed and I needed to update my biome data files for the tests to pass?

stefandollase:
I am glad you got it working.

Yes, this is normal, however I agree that it is unexpected at first: To keep the stored biome data small, the test data generator only stores the biome data which are requested during the generation of the test data. Indeed, it stores the exact requests. When the tests are executed, only the exact same requests can be answered. As a side effect of this, the stored biome data changes as soon as the requests change, which is the case when the algorithm is changed.

Treer:
Ahh, makes perfect sense. thanks.

PR #123, opened 2016-03-06
Working filtered random seed demo

mgod:
Addresses #120.

Working search configured by a search.json file in the home directory. Feature work:

- [x] Basic filter structure
- [x] Run filter correctly in ```Actions```
- [x] Build JSON interface to specify parameters for
  - [x] biome filters
  - [x] structure filters
  - [x] continuous search with logging
- [ ] finish documentation

PR #129, opened 2016-03-13
Working filtered random seed demo

mgod:
Addresses #120. See original PR #123 for more comments.

Working search configured by a search.json file in the home directory. Feature work:

- [x] Basic filter structure
- [x] Run filter correctly in ```Actions```
- [x] Build JSON interface to specify parameters for
  - [x] biome filters
  - [x] structure filters
  - [x] continuous search with logging
- [ ] finish documentation

### DIFF
--- a/src/main/java/amidst/gui/main/Actions.java
+++ b/src/main/java/amidst/gui/main/Actions.java
@@ -6,9 +6,6 @@ import java.awt.Toolkit;
 import java.awt.datatransfer.StringSelection;
 import java.awt.image.BufferedImage;
 import java.io.File;
-import java.io.FileOutputStream;
-import java.io.FileWriter;
-import java.io.PrintWriter;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
@@ -24,12 +21,12 @@ import amidst.gui.main.viewer.ViewerFacade;
 import amidst.logging.Log;
 import amidst.mojangapi.MojangApi;
 import amidst.mojangapi.file.MojangApiParsingException;
+import amidst.mojangapi.minecraftinterface.local.LocalMinecraftInterfaceCreationException;
 import amidst.mojangapi.minecraftinterface.MinecraftInterfaceException;
-import amidst.mojangapi.world.World;
 import amidst.mojangapi.world.WorldSeed;
 import amidst.mojangapi.world.WorldType;
 import amidst.mojangapi.world.coordinates.CoordinatesInWorld;
-import amidst.mojangapi.world.filter.WorldFilter;
+import amidst.mojangapi.world.filter.WorldFinder;
 import amidst.mojangapi.world.icon.WorldIcon;
 import amidst.mojangapi.world.player.Player;
 import amidst.mojangapi.world.player.PlayerCoordinates;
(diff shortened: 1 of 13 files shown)


jmessenger235:
I have had some time to test the implementation and it is really solid. There are a few things that are problematic though and I mention for consideration. 

First, this has major memory consumption issues when working with larger search distances. i.e. the quartered region search has issues when you start needing to allocate the arrays for 8192+ searches.

Second, due to the lack of internal concurrent searches, I have run multiple windows at the same time to use multiple threads to speed up searches for more complex criteria. Although this works, it seems to have some very strange silent crashing behavior. i.e. it will stop searching without any error. I have been unable to reproduce this via eclipse to debug it for some reason and need to do some more testing.

Other Considerations:
Because of the first problem, it is nearly impossible to add something like a Quad Witch hut scan criteria for the within that 8-10k range as you get too many out of memory crashes.

The Json does not seem to allow empty biome or structure filters as it seems to cause a json exception. Documentation on some common Json pitfalls would likely prevent a number of bugs.

Down the road, if the first issue can be resolved by increasing the granularity of the search, then it might be useful to allow searches of a particular criteria within a given seed. This would be useful for determining location for a base on a server where the seed is already set.

As already mentioned in the older PR, some degree of fuzzy matching would be a positive addition.

@mgod are you planning on re-basing this PR so it can be more easily merged in?
@stefandollase do you have any idea when you are planning on reviewing this for a merge?

stefandollase:
@jmessenger235
Thanks for testing it. This seems to be valuable feedback :-)

I would really like to spend more time for the project. However, I currently don't have any spare time, because I am studying and working. I tried to predict when I will be able to spend more time again, but it did not work out as expected. Thus, I won't try to predict it again. I will do stuff when I do it. Until then, I will try my best to answer the issues and react to questions in the minecraftforum.

Regarding this PR: I think this is a quite complex feature, which requires multiple PRs, maybe by different people to become releasable. Thus, we should have a [Feature Toggle](http://martinfowler.com/bliki/FeatureToggle.html) for it. The point here is, that we integrate the half-finished feature into master, but we don't expose the feature to the user. This allows us to work on the feature collaboratively, without holding back new releases from the master branch. This also prevents a long living feature branch, which always has the risk of a big merge conflict when it is finally integrated back into the master. What I mean by feature toggle here is a simple boolean flag in a new class `amidst.FeatureToggles`. This flag will be checked when creating the menu entry for this feature. We should make sure that the feature is always disabled when it is integrated into the master.

Because there will be more pull requests about this feature, we should carefully separate the discussion between the pull requests and #120. #120 should contain the conceptual discussion while each pull request should contain the discussion that is specific to its implementation.

@mgod
Sorry for letting you wait that long. I am not used to manage such a project. Also, I was hesitant to integrate the feature, because I was worried about blocking releases for the master, but I think the feature toggle is a good solution to this problem.

Can you get this pull-request conflict-free? You can merge in the current master to do so or re-base onto the current master. Once it is conflict-free I will review and integrate it soon. You can implement the feature toggle by yourself, or else I will do it after the integration.

jmessenger235:
Agreed on the separation and the need for further PRs. I will move the appropriate comments to #120 along with some other things that I plan to do a PR for when this one gets through. This is a phenomenal start though. Excellent work @mgod! 

I did more testing on the crashing problem and haven't been able to reproduce it for debugging. I did however find out my ram has at least one bad dimm so it could be hardware. The other time eclipse crashed while running overnight. In short, I'm not sure what is going on with that issue.

mgod:
I should be able to make these changes this weekend.

mgod:
@stefandollase This should now be merge-able. I didn't do a formal feature toggle, though I did change it so the menu item doesn't appear unless there is a ```search.json``` file. I'm not sure if you'll want to do a more formal toggle on top of this.

@jmessenger235 The memory issue is something I noted during development, though the ```BaseFilter``` public ```isValid``` implementation should be relatively easy to modify to use one array to get world data multiple times and see if the abstract ```isValid``` passes multiple times. I suspect this will speed up search times a bit as well, though I'd rather get this merged in and open a separate PR for that.

I'm not sure if concurrency will be easy to do – I don't quite understand the ```amidst.mojangapi.MojangApi``` class well enough to know. Hopefully someone with a better understanding can point the way.

jmessenger235:
I think that as a feature toggle would be adequate.

@mgod Thank you for that pointer. I may work on improving the memory situation over the week when I get a chance along with Quad Witch Hut searching. 

@stefandollase What do you think of just running multiple separate world filters via findRandomWorld in WorldFinder.java? Do you think there would be issues with that?

Issue #132, opened 2016-03-15
Can't compile without jnbt-1.3


mathuin:
I want to use AMIDST but I don't know much about Java and I don't want to install Java 8 on my system.  So, I checked out the source code and made a Dockerfile.  The file referenced java:8, installed maven, and attempted to run 'mvn -e package'.

This was found in the scrollback:

```
Downloading: https://repo.maven.apache.org/maven2/jnbt/jnbt/1.3/jnbt-1.3.pom
[WARNING] The POM for jnbt:jnbt:jar:1.3 is missing, no dependency information available
```

This was at the end:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 17.748s
[INFO] Finished at: Tue Mar 15 17:29:41 UTC 2016
[INFO] Final Memory: 9M/239M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project amidst: Could not resolve dependencies for project amidst:amidst:jar:0.0.1-SNAPSHOT: Could not find artifact jnbt:jnbt:jar:1.3 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project amidst: Could not resolve dependencies for project amidst:amidst:jar:0.0.1-SNAPSHOT: Could not find artifact jnbt:jnbt:jar:1.3 in central (https://repo.maven.apache.org/maven2)
```

Is the jar supposed to be in the central Maven repository?  Do you guys have it installed locally or something?  Help!  Thanks!

mathuin:
... okay, I feel a little dumb for this.

In your Travis XML file,  you call 'mvn clean'.  That apparently installs ./lib/JNBT_1.3.jar into the .m2/repository/jnbt/jnbt/1.3 directory.  It's even obliquely referenced in the comments.

Sorry about that!  I'm going to close the issue -- but hopefully if someone else has it, they'll find this in a search and solve their problem. :-)

stefandollase:
Good thing you figured it out :-) I have to admit, it is a little bit confusing, but I was not able to figure out a better solution for this. Just for the future if somebody finds this issue: We have a [wiki page](https://github.com/toolbox4minecraft/amidst/wiki/Building-from-Source-Code) which describes the build process.

[...]

PR #137, opened 2016-03-22
Dockerfile added to build jar file

mathuin:
The Dockerfile will build the jar file and copy it to a target volume.  This
allows users to compile the jar file without having Java 8 or Maven installed
on their local systems.  The resulting jar file will of course require Java 8.

To use the Dockerfile to build the binary, run the following commands:
```
docker build -t makeamidst .
docker run --rm -it -v `pwd`:/target makeamidst
```
These commands will construct a Docker container suitable for building the jar
file, and then build the jar file itself and copy it into the current directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM java:8
+RUN apt-get update
+RUN apt-get install -y maven
+RUN mkdir /target
+VOLUME /target
+COPY . /opt/amidst
+WORKDIR /opt/amidst
+RUN ["mvn", "clean"]
+RUN ["mvn", "package"]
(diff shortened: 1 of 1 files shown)


stefandollase:
Thanks for providing the Dockerfile. However, I guess I did not clearly state my intention: I did not mean to ask you to create a pull-request, but to simply paste the contents of the Dockerfile to the original issue. This is because the current project setup does not use Docker and thus adding it to the repository might confuse some people. I just thought, it might be helpful in the future if there is a need to use a technology like Docker. For now, I like to keep it simple.

I will post the Dockerfile to issue #132.

PR #145, opened 2016-04-03
Support early betas

Treer:
Minecraft's world generation system was completely replaced in beta 1.8 with new fractal-based biome system. The new fractal-based system is the one Amidst currently works with, which prevents it from being able to show early beta worlds.

### Why would anyone want to see early beta worlds? 
The oldest still-running Minecraft server in world started in December 2010 — probably around b1.1 — and to this day it remains popular, its blasted landscape is famous, and you have to travel tens of kilometers before you can find chunks not generated with the early system.

In implementing this I also discovered that the original [temperature & rainfall based](http://minecraft.gamepedia.com/File:BiomesGraph.png) world generation system is damn cool - still better in some ways than the current system.

### The old ways

The wiki has some descriptions of the [early beta world generation](http://minecraft.gamepedia.com/Biome/Before_Beta_1.8).

The biggest difference for Amidst is that there is no ocean biome  :scream:

### Issues
This pull request should be discussed before merging into the master branch, I've opened it so that discussion can take place here (and if anyone else is wants to help test or refine it).

Things to consider:
* Oceans - the map is less useful without these. Oceans are created by filling in the land up to level 64 with water blocks - there is no ocean biome. I intend to try to add oceans to the map, but there's a good chance it will be too difficult (I'm hoping to be able to generate the same noise map it uses for land height).
* Spawn - I don't know the algorithm, so the spawn point is always going to 0,0. I heard the early spawn had something to do with beaches, if that's true then it might be hard as there's no beach biome, see oceans point above. Anyone got info? 
* Impact on the rest of the app - This is an extremely minor feature for most users, so this code musn't be high maintenance or adversely impact normal operation. I believe it will be low-maintenance because the beta versions of Minecraft will never change - if the code works now then it should always work. However speed needs to be checked:
 * Due to the size of the World class, I had to [increase the max size of classes to match against](https://github.com/toolbox4minecraft/amidst/pull/145/commits/c4539ab6308b6756277af5e6962aef18fa43b5a0#diff-e9636a9ba7f0e19e0c6e358011cf7c77R19). In the past, Amidst ignored any class larger than 8k, now it scans all classes up to 32k. Most classes are small, so this probably makes no difference, but it should probably be timed.
* Automated tests - no work done on checking or updating these.
* Manual tests - I haven't tested the biome profile and colouring tools, lots of changes were made to biomes.

### Screenshot

To save you having to compile it to see the early betas

image 4

Fwooor!

### DIFF
--- a/src/main/java/amidst/clazz/real/RealClass.java
+++ b/src/main/java/amidst/clazz/real/RealClass.java
@@ -46,6 +46,7 @@ public class RealClass {
 	private final List<String> utf8Constants;
 	private final List<Float> floatConstants;
 	private final List<Long> longConstants;
+	private final List<Integer> integerConstants;
 	private final List<Integer> stringIndices;
 	private final List<ReferenceIndex> methodIndices;
 
@@ -58,9 +59,10 @@ public class RealClass {
 	RealClass(String realClassName, byte[] classData, int minorVersion,
 			int majorVersion, int cpSize, int[] constantTypes,
 			RealClassConstant<?>[] constants, List<String> utf8Constants,
-			List<Float> floatConstants, List<Long> longConstants,
-			List<Integer> stringIndices, List<ReferenceIndex> methodIndices,
-			int accessFlags, RealClassField[] fields, int numberOfConstructors,
+			List<Float> floatConstants, List<Long> longConstants, 
+			List<Integer> integerConstants, List<Integer> stringIndices, 
+			List<ReferenceIndex> methodIndices,	int accessFlags, 
+			RealClassField[] fields, int numberOfConstructors,
 			int numberOfMethods) {
 		this.realClassName = realClassName;
 		this.classData = classData;
@@ -71,6 +73,7 @@ public class RealClass {
 		this.constants = constants;
 		this.utf8Constants = utf8Constants;
 		this.floatConstants = floatConstants;
+		this.integerConstants = integerConstants;
 		this.longConstants = longConstants;
 		this.stringIndices = stringIndices;
 		this.methodIndices = methodIndices;
@@ -134,6 +137,15 @@ public class RealClass {
 		return false;
 	}
 
+	public boolean searchForInteger(int required) {
+		for (Integer entry : integerConstants) {
+			if (entry.intValue() == required) {
+				return true;
+			}
+		}
+		return false;
+	}	
+	
 	public boolean searchForLong(long required) {
 		for (Long entry : longConstants) {
 			if (entry.longValue() == required) {
--- a/src/main/java/amidst/clazz/real/RealClassBuilder.java
+++ b/src/main/java/amidst/clazz/real/RealClassBuilder.java
@@ -34,10 +34,12 @@ public class RealClassBuilder {
 			List<String> utf8Constants = new ArrayList<String>();
 			List<Float> floatConstants = new ArrayList<Float>();
 			List<Long> longConstants = new ArrayList<Long>();
+			List<Integer> integerConstants = new ArrayList<Integer>();
 			List<Integer> stringIndices = new ArrayList<Integer>();
 			List<ReferenceIndex> methodIndices = new ArrayList<ReferenceIndex>();
 			readConstants(stream, cpSize, constantTypes, constants,
-					utf8Constants, floatConstants, longConstants, stringIndices);
+					utf8Constants, floatConstants, longConstants, 
+					integerConstants, stringIndices);
 			int accessFlags = readAccessFlags(stream);
 			skipThisClass(stream);
 			skipSuperClass(stream);
@@ -50,7 +52,7 @@ public class RealClassBuilder {
 					- numberOfConstructors;
 			return new RealClass(realClassName, classData, minorVersion,
 					majorVersion, cpSize, constantTypes, constants,
-					utf8Constants, floatConstants, longConstants,
+					utf8Constants, floatConstants, longConstants, integerConstants,
 					stringIndices, methodIndices, accessFlags, fields,
 					numberOfConstructors, numberOfMethods);
 		} else {
@@ -77,13 +79,14 @@ public class RealClassBuilder {
 	private void readConstants(DataInputStream stream, int cpSize,
 			int[] constantTypes, RealClassConstant<?>[] constants,
 			List<String> utf8Constants, List<Float> floatConstants,
-			List<Long> longConstants, List<Integer> stringIndices)
-			throws IOException {
+			List<Long> longConstants, List<Integer> integerConstants, 
+			List<Integer> stringIndices) throws IOException {
 		for (int q = 0; q < cpSize; q++) {
 			byte type = stream.readByte();
 			constantTypes[q] = type;
 			constants[q] = readConstant(stream, type, utf8Constants,
-					floatConstants, longConstants, stringIndices);
+					floatConstants, longConstants, integerConstants, 
+					stringIndices);
 			if (RealClassConstantType.isQIncreasing(type)) {
 				q++;
 			}
@@ -92,13 +95,13 @@ public class RealClassBuilder {
 
 	private RealClassConstant<?> readConstant(DataInputStream stream,
 			byte type, List<String> utf8Constants, List<Float> floatConstants,
-			List<Long> longConstants, List<Integer> stringIndices)
+			List<Long> longConstants, List<Integer> integerConstants, List<Integer> stringIndices)
 			throws IOException {
 		switch (type) {
 		case RealClassConstantType.STRING:
 			return readString(stream, type, utf8Constants);
 		case RealClassConstantType.INTEGER:
-			return readInteger(stream, type);
+			return readInteger(stream, type, integerConstants);
 		case RealClassConstantType.FLOAT:
 			return readFloat(stream, type, floatConstants);
 		case RealClassConstantType.LONG:
@@ -138,8 +141,9 @@ public class RealClassBuilder {
 	}
 
 	private RealClassConstant<Integer> readInteger(DataInputStream stream,
-			byte type) throws IOException {
+			byte type, List<Integer> integerConstants) throws IOException {
 		int value = stream.readInt();
+		integerConstants.add(value);
 		return new RealClassConstant<Integer>(type, value);
 	}
 
--- a/src/main/java/amidst/clazz/real/RealClasses.java
+++ b/src/main/java/amidst/clazz/real/RealClasses.java
@@ -16,7 +16,7 @@ import amidst.documentation.Immutable;
 public enum RealClasses {
 	;
 
-	private static final int MAXIMUM_CLASS_BYTES = 8000;
+	private static final int MAXIMUM_CLASS_BYTES = 34000;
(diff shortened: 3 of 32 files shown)


stefandollase:
Wow, I did not expect that :D Great job!

It is really unfortunate that there is no Ocean biome. This means, we have to use a similar heuristic like we did for the end islands?

I guess we can live with the spawn point not being displayed correctly. It is nice to have the beta versions covered, even if some parts are still missing or incorrect. However, we should provide documentation about what is working and what is not working. But that has time until this feature is released.

I agree that the new code should be low maintenance, however I cannot yet judge about this, because I did not have a look at the code. Regarding the increased maximum class size: We should indeed do some speed testing. An alternative is to use a lower barrier by default and slowly increase the barrier if the recognition failed. This would keep the recognition quick for recent versions, but for older version it would be even slower. But I would like to avoid this complexity if the current speed is sufficient.

I did not yet look at the code too closely. I hope I get the time to do some work for the project in the next few week.

Treer:
I tried adding oceans by having minecraft generate the noise field, but it generates it at full 16 × 16 × 128 resolution for every chunk, so while Amidst was able to display oceans, it was far too slow.

I've left the ability in there, but turned it off — it can be used as a reference if we try to generate the noise field only at level 63.

image 5

Treer:
Until oceans are done, it's hard to know which extensions to the clazz hooking system will still be needed - probably most of them but we may be able to simplify `MinecraftClassTranslator_EarlyBetas.java` once we know what we don't need.

However, something I suspect will be wanted is the ability to detect a class based on it's symbolic superclass. I'll have difficulty pulling that off because I don't fully understand how the matching system works, so if you start refactoring this branch in that area (and I don't expect you to), consider ways to implement that functionality. 

Here's the problem:

Early betas up to b1.5 look like this:
```
class Dimension {
    // assumes the dimension is Overworld
    <All the good stuff we need> 
    ⁝
}

class NetherDimension extends Dimension {
    // overrides some stuff to make it more nethery
    ⁝
}
```

As of b1.6  it looks like this
```
abstract class Dimension {
    // assumes the dimension is Overworld
    <All the good stuff we need> 
    ⁝
}

class OverworldDimension extends Dimension { } // there is literally nothing inside the braces

class NetherDimension extends Dimension {
    // overrides some stuff to make it more nethery
    ⁝
}
```

So by making Dimension abstract, we're left with no way to instantiate it, and no way to locate `OverworldDimension ` because it's empty - there's nothing to match against.

So, I think oceans will be possible, but only for versions before b1.6 unless there's a way for Amidst to instantiate either the abstract `Dimension` or the empty `OverworldDimension`.

In the worst case scenario, I'm OK with b1.6 and b1.7 missing oceans.

stefandollase:
You might be able to match against the number of constructors/methods/fields to be `0`, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/clazz/translator/CTBuilder.java#L82). However, this is just a wild guess and might fail entirely.

It seems to be reasonable to search for a class by its superclass. Did you figure out how to extract the super-class from the byte-code? Currently, we [skip the super-class](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/clazz/real/RealClassBuilder.java#L43). If we can extract the real-class name, it might be possible to create a matcher against it. However, I am not entirely sure about this, because we want to specify the symbolic class name in the class translator and the mapping from symbolic class name to real class name is only available after the matchers are done. Thus, I'm afraid this will complicate the process of class recognition as a whole. But again, I am not sure about this. I will have to look at this more closely.

Treer:
I think matching against 0 constructors/methods/fields will pull up random empty interfaces, but I should doublecheck.

Since the beta versions will never change, we could take a different route and hard-code the real class name for the superclass for each version of b1.6 and above.

However, the Launcher provides 10 versions between b1.6 and b1.7.3, so RecognizedVersion would have to be able to differentiate all of them (I think it might do a better job of differentiating the earlier versions), and the class translator would need to adjust depending on the recognized version.

stefandollase:
Yes, you might be correct about the interfaces. We will have to try that. Did you know that the devtool `MinecraftVersionCompatibilityChecker` checks that there is exactly one match for each required class? So you can use that to ensure there is no more than one match :-)

Next, the class recognition is split into multiple phases, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/clazz/Classes.java#L26):

1. The load phase loads all classes from the Minecraft jar file with a size below a certain threshold. For each of these classes, a real class is created.

  The product of this phase is the list of all real classes.

2. The detect phase uses a combination of real class detectors to find each real class that matches the given requirements. Here, usually all detectors have to match. However you can also use the `.or()` method to split a list of detectors. In this case, it is enough if all detectors before the `.or()` are matching **or** if all detectors after `.or()` are matching. You can also use `.or()` multiple times.

  The product of this phase is a mapping from the symbolic class declaration to the real class name.

3. The declare phase loads the `Class` objects from the Minecraft jar file, which have been identified in the previous phase. Afterwards, it creates a `SymbolicClass` wrapper for each of these loaded classes. It also creates wrappers for all the methods, constructors and fields that should be available in the `LocalMinecraftInterface`.

  The product of this phase is a mapping from the symbolic class name to the fully constructed symbolic class object.

Please note that the class translator as well as all of its parts, namely the real class detectors and symbolic class declarations, is stateless and can thus be reused. It is only used during the recognition process.

The real class is also only used during the recognition. It represents one specific class in the Minecraft jar file. Also, it provides all functions to recognize a specific class. Every real class detector basically calls some methods on the real class and returns a boolean based on the result.

The symbolic class is created during the recognition process and used by the `LocalMinecraftInterface` to actually interact with the Minecraft code. Since symbolic classes can reference each other, the creation of the symbolic classes is quite complex. For example the symbolic method initialize all biome generators of the symbolic class gen layer refers to the symbolic class world type, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java#L45). (BTW, I just noticed that the String `"WorldType"` should be replaced by `SymbolicNames.CLASS_WORLD_TYPE`. I wrote this down as a to do for myself :-P). Because of this graph-like structure, there is a class called [`SymbolicClassGraphBuilder`](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/clazz/symbolic/SymbolicClassGraphBuilder.java).

The `Class` object that is wrapped by the symbolic class is how Java works with its byte code. Java cannot do anything with the information provided by a real class object. The `Field`, `Method` and `Constructor` instances in the symbolic field, method an constructor instances work analogous.

The take-away from all of this is, that after the detect phase is done, the class recognition is effectively done. The declare phase exists only to actually make members of the detected classes available to be used in the `LocalMinecraftInterface`. This is also the reason for the devtool `MinecraftVersionCompatibilityChecker` to skip the declare phase: It only checks if all required classes have an exact match in all Minecraft versions. It does not care about the methods, constructors and fields that are used by the `LocalMinecraftInterface`.

To enable a detector to make use of the symbolic class name, we have to adjust the class translator, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/clazz/translator/ClassTranslator.java#L43). This algorithm needs to be adjusted, so it first detects all classes whose detector does not depend on other classes. Then it detects all classes whose detector does only depend on already detected classes and so on. This might work, as long as there is no cyclic dependency between the real class detectors ... the algorithm has to be able to start somewhere. Nevertheless, I still would like to omit this complexity, if possible.

Regarding your suggestion to hard code the real class names for the early beta versions: I also thought about doing this, not only for the early beta versions, but for all versions. The mapping from real class to symbolic class would simply be hard-coded for all the versions. We could use the version features mechanism to select the correct mapping. We could also provide a JSON document, with the hard-coded mapping, just like we provide the update document via Github Pages. Of course, we would also deliver the latest version of this document with each release. This allows us to fully omit the complexity of the class recognition mechanism from Amidst.

However, there is a downfall to this: We currently cannot uniquely identify a specific version. We basically have to hope for each new Minecraft version that the magic string was changed. An alternative to the magic string is to use a check-sum of the Minecraft jar file to uniquely identify it. This will probably work most of the time, but it completely closes the door to ever support any modded Minecraft version. Another alternative is to use the [selected Minecraft version](https://github.com/toolbox4minecraft/amidst/wiki/FAQ#what-is-all-the-information-in-the-window-title) directly, however this removes the ability to use Amidst by only providing a Minecraft jar file via the command line. Nevertheless, I think we should think about switching the version recognition to something more accurate and reliable. Maybe, I should open a new discussion issue for this.

After all, I am a bit undecided about what is the best course of action. I would really like to hear your opinion about that.

BTW: You can declare the class translator itself as a version feature. This allows you to use different class translators for different versions :-)

Treer:
I rather like this idea... whenever the list contains all the requested symbolic class info, Amidst could entirely skip phase 1 and phase 2, which could make switching profiles an almost instant operation.

Plus I assume the MinecraftVersionCompatibilityChecker devtool could be extended to build the list automatically.

The magic version string is [massive and unique for the early versions of Minecraft](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/minecraftinterface/RecognisedVersion.java#L94), so I'm guessing it was designed with them in mind somehow and we might be able to adjust it to provide fuller coverage of all the versions — I don't like the idea of using a hash either.

> Of course, we would also deliver the latest version of this document with each release.

Just to be clear, you mean inside the Amidst .jar, right? I don't mind if it looks for an external JSON file to update/override the data in the .jar (that could be handy), but it should contain the list without needing the internet or any extra files.

And as you say, it removes any need for making phase 2 iterative, which would probably add more complexity than the payoff would be worth.

Treer:
We can use the hashes in the META-INF/MANIFEST.MF, since modded versions must remove this file. 

(I edited this into my previous post but wasn't sure if you'd already read it)

if the manifest is missing, or unrecognised, Amidst can drop back to using class detection. 

This method wouldn't allow oceans in some modded betas, due to the limitations of class detection for b1.6, but I think we can all still sleep at night with this limitation.

stefandollase:
I created a discussion issue about this topic to prevent this pull-request from getting even more off-topic, see #146. Maybe, we need to resolve that new issue first, before we can decide about how to handle the class recognition problem in this issue.

The current magic string calculation does not require phase 1 to run. It loads a specific class and uses reflection to determine the type of the fields of that class. After some more string replacements, it produces the magic string, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/minecraftinterface/RecognisedVersion.java#L124).

Yes, I meant to include the latest version of the JSON document in the Amidst jar file. Additionally, if an internet connection is available Amidst would download more recent information via Github Pages.

I don't actually know how Java handles signed code. Thus, I also don't know which files have to be changed/removed by modded Minecraft. However, I am also quite sure that the signature is created from the obfuscated classes, since Java can only verify the signature against the obfuscated classes.

Treer:
I'm going to use this branch as a development branch for the feature. When it's finished I'll merge the changes I want to keep into a new branch (in logical bundles) and made the pull request from that. So I'm closing this pull request, but here is still where any discussion will take place.

Issue #134, opened 2016-03-17
The keyboard shortcut to zoom out is broken for some keyboard layouts.


MOARdV:
I don't have a mouse wheel to control zooming.  I have discovered that the `-` key on the keyboard will zoom out, but I haven't found a way to zoom in (the `=` does not, although that would be my expectation).  I may have overlooked the keyboard shortcuts documentation, but I don't see a way to zoom back in using keyboard controls.

stefandollase:
Yes, we should add documentation of the keyboard shortcuts, this is correct. I will add this to #66.

To answer your question: The zoom in key is `+`.  I guess `=` makes sense on the English keyboard layout? However, it does not make sense on most of the other keyboard layouts. Thus, I decided to use `+`, because it makes semantically sense.

MOARdV:
The `=` on the keyboard does not zoom (US/English keyboard - the buttons are located to the right of the row of numbers).  I tried Shift + `=` (the `+` character), and it likewise does not zoom.  This keyboard does not have the 10-key numeric pad, since it is like a laptop keyboard.  I can take a look at the amidst code and see if I can provide a suggestion - I don't speak Java natively, so I haven't looked at this code.

stefandollase:
Okay, I think I have an explanation for this. The old Amidst used `=` to zoom in. As I said above, I switched it to `+`, so it makes semantically sense. However, when I changed that, I also noticed that the `=` key does not work for me. I did not further think of this, because the `+` key was working. However, now I have an explanation for this: For me, die `=` character is typed using **Shift + 0**. For you, the `+` character is typed using **Shift + =**. This means, the code that catches the zoom keys does not interpret the pressed keys as characters but only as key presses. At least this is my hunch on this. I will look into this before the next release.

If you want to look into the code by yourself, here are the important lines:

* [zoom keys in the old code](https://github.com/skiphs/AMIDST/blob/master/src/MoF/MapViewer.java#L366)
* [zoom keys in the new code](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/gui/main/MainWindow.java#L124)

MOARdV:
Okay, thank you.  I will see if I can find out how VK_PLUS maps on my keyboard.

stefandollase:
@MOARdV I just created a new pre-release which contains a fix for this issue by @MtnViewJohn.

Thus, I close this issue.

PR #157, opened 2016-04-19
Make zoom in/out key events work for all keyboard layouts

MtnViewJohn:
I am affected by Issue #134 and I suggest using the keyTyped event instead keyPressed. This way + and - will work for zoom no matter how they are generated by the user's keyboard (even the keypad).


stefandollase:
Thanks for fixing it. It will be part of the next release. I will create a pre-release right now, so this is no longer an issue.

PR #167, opened 2016-05-08
Add better naming for BiomeType class.

moulins:
I was browsing the source code and I saw that this class did not have clear names.

The names of the types are not taken from the game ; they are guesses based on the biomes they contain, but I think they are accurate.


stefandollase:
Thanks, this was indeed one of the open TODOs. It looks good. I will merge it this week.

Issue #205, opened 2016-07-12
Friendly error message for Optifine and potentially other incredibly popular mods


ekevoo:
The amount of bug reports including Optifine is impressive.

![Too damn high!](https://imgflip.com/s/meme/Too-Damn-High.jpg)

Maybe before crashing and burning with a huge stack trace that only the initiated can understand, we should look at things like profile name and library names for, say, common bad words? :)

stefandollase:
It is impressive, indeed! This is definitely going into the 4.2 release :-P

ArgTang:
@stefandollase Could i get feedback on the pull rewuest for this issue?

stefandollase:
@ArgTang
I reviewed your pull request. The idea should work, but there are still some rough edges that need to be worked out.

A more elaborate fix would depend on the changes that I proposed [here](https://github.com/toolbox4minecraft/amidst/issues/146#issuecomment-208092539). However, I did not yet get the time to implement these changes. They would allow us to use the vanilla Minecraft version that the modded Minecraft version is based on.

stefandollase:
#218 implements this feature

Issue #151, opened 2016-04-12
Amidst keeps crashing.


glokuosta:
I was never able to run amidst for once in my life. Neither on this computer or my old one. I used old versions, newest version, another thing named amidst exporter, but they all gave me errors. 

```
[info] Amidst v4.1
[info] Current system time: 2016-04-12 23:19:22.405
[info] System.getProperty("os.name") == 'Windows 10'
[info] System.getProperty("os.version") == '10.0'
[info] System.getProperty("os.arch") == 'amd64'
[info] System.getProperty("java.version") == '1.8.0_77'
[info] System.getProperty("java.vendor") == 'Oracle Corporation'
[info] System.getProperty("sun.arch.data.model") == '64'
[info] Not using OpenGL.
[info] Biome profile activated: default
[info] using '.minecraft' directory at: 'C:\Users\X\AppData\Roaming\.minecraft', libraries: 'C:\Users\X\AppData\Roaming\.minecraft\libraries'
[info] using biome profiles at: 'biome'
[info] Scanning for profiles.
[info] Successfully loaded profile list.
[info] Beginning latest version list load.
[info] Attempting to download remote version list...
[info] Successfully loaded version list. URL: https://launchermeta.mojang.com/mc/game/version_manifest.json
[warning] cannot find valid version directory for launcher profile 'Forge'
[warning] cannot find valid version directory for launcher profile '1.8-ShadersMod'
[warning] cannot find valid version directory for launcher profile '1.7 forge'
[info] using minecraft launcher profile '1.9' with versionId '1.9.2'
[info] Loading libraries.
[info] Found library: C:\Users\X\AppData\Roaming\.minecraft\libraries\oshi-project\oshi-core\1.1\oshi-core-1.1.jar
[info] Found library: C:\Users\X\AppData\Roaming\.minecraft\libraries\net\java\dev\jna\jna\3.4.0\jna-3.4.0.jar
[info] Found library: C:\Users\X\AppData\Roaming\.minecraft\libraries\net\java\dev\jna\platform\3.4.0\platform-3.4.0.jar
[info] Found library: C:\Users\X\AppData\Roaming\.minecraft\libraries\com\ibm\icu\icu4j-core-mojang\51.2\icu4j-core-mojang-51.2.jar
[info] Found library: C:\Users\X\AppData\Roaming\.minecraft\libraries\net\sf\jopt-simple\jopt-simple\4.6\jopt-simple-4.6.jar
[...]

PR #218, opened 2016-07-29
Message when trying to load modded profiles

ArgTang:
Hi

This will let users know that amidst need a vanilla profile, and returns to the profileselector afterwards, when trying to select a modded profile.

This is a small fix, and will probably not catch all modded profiles, but most of them. 
To get a better fix, we would have to check each profile with the profile list that we get from mojang, And mark all profiles that do not mach as modded and display some sort of message. I am not shure where or how to do this, but i can look into it if need be.

Let me know what you think.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = tabs
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,0 @@
-# EditorConfig is awesome: http://EditorConfig.org
-
-# top-most EditorConfig file
-root = true
-
-[*]
-indent_style = tab
-indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,5 +4,5 @@
 root = true
 
 [*]
-indent_style = tabs
+indent_style = tab
 indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -5,11 +5,9 @@
 *.ear
 
 # IntelliJ IDEA project files #
-*.iml
-/.idea/
-/out/
-
-# Netbeans files
+./AMIDST.eml
+./AMIDST.iml
+./.idea/*
 /bin
 /dist
 /logs
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,11 @@
 *.ear
 
 # IntelliJ IDEA project files #
-./AMIDST.eml
-./AMIDST.iml
-./.idea/*
+*.iml
+/.idea/
+/out/
+
+# Netbeans files
 /bin
 /dist
 /logs
--- a/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
+++ b/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
@@ -84,7 +84,7 @@ public class LocalProfileComponent extends ProfileComponent {
 		try {
 			Log.i("using minecraft launcher profile '" + getProfileName() + "' with versionId '" + getVersionName() + "'");
 
-			String possibleModProfiles = ".*[optifine|forge].*";
+			String possibleModProfiles = ".*(optifine|forge).*";
 			if (Pattern.matches(possibleModProfiles, getVersionName().toLowerCase(Locale.ENGLISH))) {
 				Log.e("Amidst does not support modded Minecraft profiles! Please select or create an unmodded Minecraft profile via the Minecraft Launcher.");
 				return false;
--- a/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
+++ b/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
@@ -81,6 +81,13 @@ public class LocalProfileComponent extends ProfileComponent {
 		try {
 			Log.i("using minecraft launcher profile '" + getProfileName() + "' with versionId '" + getVersionName()
 					+ "'");
+					
+			String possibleModProfiles = "optifine|forge";
+            if (Pattern.matches(possibleModProfiles, getProfileName().toLowerCase())) {
+                Log.e("Amidst does not support modded Profiles, please select or create an unmodded profile");
+                return false;
+            }
+
 			mojangApi.set(getProfileName(), profileDirectory, versionDirectory);
 			return true;
 		} catch (LocalMinecraftInterfaceCreationException e) {
--- a/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
+++ b/src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
@@ -1,7 +1,5 @@
 package amidst.gui.profileselect;
 
-import java.io.FileNotFoundException;
-
 import amidst.Application;
 import amidst.documentation.AmidstThread;
 import amidst.documentation.CalledOnlyBy;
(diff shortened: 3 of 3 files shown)


ArgTang:
#205

.editorconfig
```diff
@@ -0,0 +1,8 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = tab
+indent_size = 4
```
stefandollase: Please don't make changes that are unrelated to the feature.

.gitignore
```diff
@@ -5,9 +5,11 @@
...
 
 # IntelliJ IDEA project files #
-./AMIDST.eml
-./AMIDST.iml
-./.idea/*
+*.iml
+/.idea/
+/out/
+
+# Netbeans files
```
stefandollase: Please don't make changes that are unrelated to the feature.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
```diff
@@ -1,7 +1,5 @@
 package amidst.gui.profileselect;
 
-import java.io.FileNotFoundException;
```
stefandollase: Please use Eclipse to prevent changing the order of import statements.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
```diff
@@ -81,6 +82,13 @@ private boolean tryLoad() {
 		try {
 			Log.i("using minecraft launcher profile '" + getProfileName() + "' with versionId '" + getVersionName()
 					+ "'");
+					
+			String possibleModProfiles = "optifine|forge";
+            if (Pattern.matches(possibleModProfiles, getProfileName().toLowerCase())) {
```
stefandollase: `Pattern#matches` requires the complete profile name to match the pattern. Thus, this only works if the complete profile name is either the word "optifine" or "forge", without anything else. For example "OptiFine 1.8.9" does not work. My suggestion is to simply use a list of strings and check if the profile name `contains` any of the strings.

`String#toLowerCase` depends on the default `Locale` on the users computer. This is an issue, see [here](https://github.com/toolbox4minecraft/amidst/issues/151#issuecomment-211903792) and [here](http://stackoverflow.com/questions/11063102/using-locales-with-javas-tolowercase-and-touppercase). However, it should work to use `.toLowerCase(Locale.ENGLISH)`.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
```diff
@@ -81,6 +82,13 @@ private boolean tryLoad() {
...
+                Log.e("Amidst does not support modded Profiles, please select or create an unmodded profile");
```
stefandollase: I propose the error message: "Amidst does not support modded Minecraft profiles! Please select or create an unmodded Minecraft profile via the Minecraft Launcher."

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
```diff
@@ -81,6 +82,13 @@ private boolean tryLoad() {
...
+                return false;
+            }
```
stefandollase: The if-block uses spaces instead of tabs for its indentation. Please use Eclipse and import the preference file that is included in this repository to ensure that your commits adhere to our coding style.

stefandollase:
Sorry for the delay. There was no evil intent, I just forgot about it after shortly looking into it when you posted it.

This should be a quick fix for the problem. Thanks! Nevertheless, there still need to be some changes until this can be merged. I marked them in the review.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
stefandollase: One more thing: Please don't check against the profile name, but against the version name. The profile name is chosen by the user and can thus be something like "1.8.9-without-optifine". The current approach would block this profile. Using the version name will prevent this issue.

stefandollase:
Also, please add the following comment in front of the introduced check:

```
// TODO: Replace with proper handling for modded profiles.
```

ekevoo:
Requiring a specific IDE is a serious deterrent to get further contributors to any Open Source project. I was considering brushing my decades-outdated knowledge of Java and maybe start contributing, but not being able to use my IDE of choice (which most likely would be IntelliJ Idea since I'm a very satisfied IntelliJ PyCharm user) is a deal-breaker.

.gitignore
ArgTang: Ill make a separate pull request for this

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
ArgTang: This is IDEA's auto reordering. ill try to manually move it back.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
ArgTang: Yes that is badly tested by me, ill fix the regex
I was not aware of the locale.

ArgTang:
@stefandollase This should adress the issues you brought up

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
stefandollase: Yes. IDEA and Eclipse both have the *organize imports* feature, however they seem to choose a different ordering. This is why I prefer to have all developers use a single set of tools. This is just an explanation, I saw you already fixed it for this pull request.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
```diff
@@ -78,9 +80,16 @@ public void load() {
...
 	@CalledOnlyBy(AmidstThread.WORKER)
 	private boolean tryLoad() {
+		// TODO: Replace with proper handling for modded profiles.
 		try {
-			Log.i("using minecraft launcher profile '" + getProfileName() + "' with versionId '" + getVersionName()
-					+ "'");
+			Log.i("using minecraft launcher profile '" + getProfileName() + "' with versionId '" + getVersionName() + "'");
+
+			String possibleModProfiles = ".*[optifine|forge].*";
+			if (Pattern.matches(possibleModProfiles, getVersionName().toLowerCase(Locale.ENGLISH))) {
```
stefandollase: This is still broken. `[]` is used to represent a set of characters. For example `[hello]` means *a single character from the set {'h', 'e', 'l', 'o'}*.

Please use the following statement:

```
Pattern.compile("optifine|forge").matcher(getVersionName().toLowerCase(Locale.ENGLISH)).find()
```

If you want to, you can extract `Pattern.compile("optifine|forge")` as a constant.

stefandollase:
@ArgTang
This is already better, however still not perfect. I added a comment.

@ekevoo @ArgTang
I see your point. However, as seen in this pull request, there are also good reasons to enforce a single set of tools for a project. Different tools often don't work that well together. However, lets not spam this pull request with a detailed discussion about it. If you want to further discuss this and want to try to change my opinion, please create a new discussion issue.

src/main/java/amidst/gui/profileselect/LocalProfileComponent.java
ArgTang: wow, yeah. Thanks for your patience. I changed the regex again.

stefandollase: Great, this should also work.

Issue #247, opened 2016-11-11
Crash opening 1.11-pre1 world


mattmoss:
```
[info] Amidst v4.2-beta2
[info] Current system time: 2016-11-11 13:18:29.263
[info] System.getProperty("os.name") == 'Mac OS X'
[info] System.getProperty("os.version") == '10.10.5'
[info] System.getProperty("os.arch") == 'x86_64'
[info] System.getProperty("java.version") == '1.8.0_91'
[info] System.getProperty("java.vendor") == 'Oracle Corporation'
[info] System.getProperty("sun.arch.data.model") == '64'
[info] Enabling OpenGL.
[info] Biome profile activated: default
[info] using '.minecraft' directory at: '/Users/mattmoss/Library/Application Support/minecraft', libraries: '/Users/mattmoss/Library/Application Support/minecraft/libraries'
[info] using biome profiles at: 'biome'
[info] Scanning for profiles.
[info] Successfully loaded profile list.
[info] Beginning latest version list load.
[info] Attempting to download remote version list...
[info] Successfully loaded version list. URL: https://launchermeta.mojang.com/mc/game/version_manifest.json
[info] using minecraft launcher profile 'mattmoss' with versionId '1.11-pre1'
[info] Loading libraries.
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/com/mojang/netty/1.6/netty-1.6.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/oshi-project/oshi-core/1.1/oshi-core-1.1.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/net/java/dev/jna/jna/3.4.0/jna-3.4.0.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/net/java/dev/jna/platform/3.4.0/platform-3.4.0.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/com/ibm/icu/icu4j-core-mojang/51.2/icu4j-core-mojang-51.2.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/net/sf/jopt-simple/jopt-simple/4.6/jopt-simple-4.6.jar
[info] Found library: /Users/mattmoss/Library/Application Support/minecraft/libraries/com/paulscode/codecjorbis/20101023/codecjorbis-20101023.jar
[...]

Issue #233, opened 2016-10-03
1.11 Woodland Mansions


MattSturgeon:
New rare structure has been added to Roofed Forest biomes in 1.11, the Woodland Mansions.

It would be good to add a layer for these.

---

As a sidenote, I wonder if looking into the implementation for the new `/locate` command will make finding structures easier?

stefandollase:
Thanks for contributing to the project! Indeed, we should add the new structure.

Regarding the `/locate` command, I just added #237.

LockManipulator:
There still doesn't seem to be a feature to find woodland mansions.

PR #269, opened 2016-11-17
Fix for #247

moulins:
This should allow Amidst to work with minecraft 1.11.

I'm not 100% sure that the fingerprints for the 2 new minecraft classes needed are future-proof, but for now it works!

I've also taken this opportunity to simplify the workings of the class translator by removing the ```RCDBuilder``` and ```***RCD``` classes (which are useless now that we have lambda-expressions).


mathuin:
While multiple anecdotes are not data, I can tell you that I was able to successfully build the JAR file from this pull request and run it with a 1.11 profile, and I'm seeing what I expect to see.

the-ntf:
It took me several hours to to get the build process right, but I was ultimately able to aply this fix and can attest that it also works for me in 1.11

stefandollase:
Thanks for finding the necessary classes and methods and for fixing the issue!

I am not yet sure whether I like that you replaced a part of the detection language my lambda expressions, but I guess it works :-P

Fear837:
Is there a version planned where you can find the new Woodland Mansions?

stefandollase:
@Fear837
See #233. We plan to support it at some unknown point in the future.

PR #316, opened 2017-05-02
Retrieve Steve's head instead of returning null

mrihtar:
In case of not being able to retrieve player's head by name or skin url amidst opens an error window. With this fix a default Steve's head is tried first before it returns null. I guess we could also use supplied "player.png" instead.

### DIFF
--- a/src/main/java/amidst/mojangapi/file/json/PlayerInformationRetriever.java
+++ b/src/main/java/amidst/mojangapi/file/json/PlayerInformationRetriever.java
@@ -18,7 +18,6 @@ public enum PlayerInformationRetriever {
 	;
 
 	private static final String SIMPLE_PLAYER_SKIN_URL = "http://s3.amazonaws.com/MinecraftSkins/";
-	private static final String STEVE_SKIN_URL = "http://assets.mojang.com/SkinTemplates/steve.png";
 
 	public static PlayerJson tryGetPlayerJsonByName(String name) {
 		try {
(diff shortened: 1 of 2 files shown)


stefandollase:
Thanks for finding the bug and submitting a fix :-)

Returning null does indeed seem to be an issue, because we don't check for null were it is used, [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/player/PlayerInformation.java#L63) and [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/player/PlayerInformation.java#L70). However, the `PlayerInformationRetriever` is not responsible to provide a default skin. This is already done [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/player/PlayerInformation.java#L20) and [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/player/PlayerInformation.java#L36). Instead, we should make sure that the available fallback cases are actually executed, e.g. by propagating the null value.

In general, we should consider switching these nullable values to have an `Optional<T>` type.

If you want to, you can make the proposed changes. Otherwise, I will fix it by myself for the next release.

mrihtar:
I opted for checking of null values in PlayerInformation for now :-).

stefandollase:
Great, thank you! I will merge it when I prepare the next release.

Issue #146, opened 2016-04-05
Rethinking the Minecraft version recognition, class recognition and usage code


stefandollase:
This issue contains the discussion about whether and how to adjust the interaction with the Minecraft jar file. Feel free to participate!

## Current steps

1. **Recognising the Minecraft version** Currently, we do this by creating a so called magic string, which is extracted via reflection from the types of the fields of a specific well known class from the Minecraft jar file, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/minecraftinterface/RecognisedVersion.java#L118).

2. **Recognising a set of classes in the Minecraft jar file** The Minecraft jar file contains obfuscated classes. This means, the classes have names like `a` or `xy`. Of course, Mojang works with proper names in their source code. The obfuscated class names are automatically generated. This means that if Mojang introduces or removes classes, there is a good chance that all other obfuscated class names are changed. This is the problem that needs to be solved by this step: We have to somehow identify the classes that we need to use from the Minecraft jar file. Currently, we do this by defining a set of requirements that are only fulfilled by the used classes, see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java#L22). This linked class translator currently works for all supported Minecraft versions. I have to say, that I am really surprised that it works so consistent across all the version since Minecraft 1.0, but it works. This step translates the obfuscated class names, which change between the different Minecraft versions, to a symbolic class name, which is basically made up for Amidst. The same class in different Minecraft versions is always mapped to the same symbolic class name.

[...]

PR #333, opened 2017-06-02
Added minecraft 1.12 pre7 support

stefandollase:
As opposed to my commit message in f8789662511c480726daff724f16271c494c20ad

>Minecraft 1.12-pre3 seems to contain malformed class files. Not only Amidst, but also standard Java decompilers cannot read them, e.g. chv.class.
>
>This change allows Amidst to ignore these classes as long as all required classes are found.

the new snapshots do not contain malformed class files. The *malformed* class files simply contain lambda expressions which we were not able to recognize (just like the decompiler that I used).

This PR also adds support to read byte code that contains lambda expressions. However, it does not allow us to search for a class file by a given lambda expression.


BigAlanM:
http://i.imgur.com/nHS8w5x.jpg
Both v4.3 beta versions seem to have a problem with PCMCv1.11.2
Interestingly v1.11.2 came out on 16w51 NOT 17w06

stefandollase:
Yes, you are right about this. Sadly, our recognition mechanism becomes more and more fuzzy, which was one of the reasons for #146. #331 should be a first step to adjust this mechanism, however it is not yet done.

See [here](https://github.com/toolbox4minecraft/amidst/blob/bc0404d20fa074ca3ffdecd768b2d5a4dd0e74aa/src/main/java/amidst/mojangapi/minecraftinterface/RecognisedVersion.java#L35) for which versions are currently grouped together by the recognition mechanism (scroll to the right). The marked line is the version you are talking about.

However, for this version, it is really only an issue because it might confuse the user. Other than that, it has no consequences. We still use the correct biome generator. The recognised version for example affects which algorithm is used for the structure generation, like temples. Thus, since we did not adjust these algorithms for newer versions of the game, it won't currently make a difference which version was recognised (see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/versionfeatures/DefaultVersionFeatures.java)).

PR #343, opened 2017-08-31
Reuse base biome colors for biome variants

moulins:
A small change to the `Biome` class; all biome variants (except Ice Plains Spikes) use the same color as their base; so there is no need to re-specify it.

(Are you certain we need to regenerate the Biome Color Table ? The special case of the Ice Plains Spikes is handled, so no biome color is modified.)


stefandollase:
Again, great work =)

stefandollase:
You are right, we don't need to regenerate the biome color table.

Issue #364, opened 2018-01-16
NullPointer exception in getGeneratorOptions when loading snapshot 18w02a


GladeGnom:
Using amidst version 4-3-beta2. Tried runing with openjre8 and openjre9.

Reproduce: run amidst, choose an mc profile with snapshot 18w02a, select 'new from seed', enter seed value and press ok. The program crashes with a nullpointer exception before showing the map.

Note that the problem so far is propably limited to snapshot 18w02a - I've tested 1.12.2, 17w43b, 17w50a amd 18w01a without problems.

below is (part of) the log and crash report:

```
[info] Unable to recognise Minecraft Version with the magic string "zjvorznfwa[Lro;tn[J[[Jtjsvstbnaqcuyoibleofog".
[info] Reading 18w02a.jar
[info] Jar load complete.
[info] Searching for classes...
[info] Found class WorldType: ask
[info] Found class BlockInit: nv
[info] Found class IntCache: bix
[info] Found class ChunkProviderSettings: bek
[info] Found class ChunkProviderSettingsFactory: aak
[info] Found class GenLayer: biz
[info] Class search complete.
[info] Loading classes...
[info] unable to find the real class method biz.a(long, @WorldType) -> GenLayer.initializeAllBiomeGenerators2
[info] unable to find the real class method biz.a(long, @WorldType, String) -> GenLayer.initializeAllBiomeGenerators3
[info] Classes loaded.
[info] Minecraft load complete.
[info] Creating world with seed '1240625079079197110' and type 'Default'
[info] Using the following generator options: 
[13:18:01] [AWT-EventQueue-0/INFO]: [STDOUT]: [crash] Amidst has encounted an uncaught exception on the thread Thread[AWT-EventQueue-0,6,main]
[13:18:01] [AWT-EventQueue-0/INFO]: [STDOUT]: [crash] java.lang.NullPointerException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at amidst.clazz.symbolic.SymbolicMethod.invoke(SymbolicMethod.java:66)
[...]

PR #366, opened 2018-01-17
fix ChunkProviderSettings class not found in 18w02a

GladeGnom:


stefandollase:
Thanks =)

closes #364

Issue #363, opened 2018-01-14
Woodland Mansions !


Divy1211:
please add the support for displaying woodland mansions in the world as well !!

stefandollase:
This is a duplicate of #233

Issue #358, opened 2017-12-21
Add a list of woodland Mansions


Ree11:
Shouldn't be to hard and would really make this app perfect for finding structures and seeds.

just like the list of end portals that can be showed by going to world and clicking stronghold have the option to jump to woodland mansions would be so nice

stefandollase:
This is a duplicate of #233

Issue #177, opened 2016-05-30
Support for Minecraft 1.10 villages


stefandollase:
Minecraft 1.10 adds villages to more biomes. We should add this to Amidst.

BigAlanM:
Actually, that should just be a toggle to turn off the part of the current Village mapping that filters out those that are too close to biome edges when using a post 1.9.4 profile and also adding the new permitted biomes.

stefandollase:
@BigAlanM 
You might be right. I will look into this when I get the time for it, which is currently not the case.

BigAlanM:
Is it possible to get a V4.2 Beta3 build out with this implemented?
There are a LOT of people on the forums that REALLY want closer to accurate PCMCv1.10 maps.

stefandollase:
I guess they will have to wait. I am currently writing my semester exams, so I have to learn. Also, I am already putting time (which I don't really have) into the project with #120. I currently prefer #120, because there are more developers working on it, so I don't want to block them.

BigAlanM:
It's been over two months since PCMCv1.10 came out (June 8, 2016).
Is there NO ONE with the knowledge and skill to tweak Amidst to accurately (or at least close) map villages in 1.10?
When this repository was started I was overjoyed that someone was going to take over for the beleaguered original programmer.
All I've seen is people hung up on expanding the functions of the program and absolutely NO support for the thousands upon thousands of loyal users that expected more timely updates.
I'm truly disappointed with you all.

ForestFeather:
This project is free, done by volunteers.  There is no real way to 'easily' determine this information, as the code that there is access to is in fact obfuscated to make understanding it harder.  Insulting the developers who do take the time to try and figure it out makes them less likely to do so.  So please don't try to act like a disappointed manager or worse, disappointed parent, and don't make claims that you speak for others.

If the issue is that important to you, you can research the algorithms that are used yourself to help provide the information to those whom may not have the skill/time but still have the programming experience, or even implement it yourself if you can code.  This is the joy of open source programs!

BigAlanM:
There were over a half a MILLION downloads of the last version of Amidst that was on Bitbucket.
Unfortunately Github doesn't record the number of downloads.
So I'm pretty sure that I do speak for, at least, thousands of Amidst users.
[...]

PR #370, opened 2018-02-21
Add woodland mansions

### LINKED ISSUE #233: 1.11 Woodland Mansions

moulins:
This adds Woodland Mansions to amidst for versions since `16w43a` (per the wiki, mansions were added in `16w39a`, but this version isn't in amidst's recognized versions list).

The constants were taken directly from the `18w05a` source code, and I've tested it in Minecraft 1.11.2.
Mansions manually found in amidst appear ingame, and mansions found ingame via the `/locate` command are shown in amidst, so it should be accurate.

I've quickly made an icon, but I think it looks too much like the Nether Fortress icon; feel free to make a better one :wink:.

Closes #233, #358, #363.


stefandollase:
Thanks for your work. This works flawlessly.

BigAlanM:
Thanks for the Mansion addition.
Unfortunately beta-3 still doesn't show all villages in the later PCMC versions.
https://www.minecraftforum.net/forums/minecraft-java-edition/seeds/2894644-flower-forest-village-thats-built-over-a-river
Seed "4927035713328246572"
Has a village close to spawn that is not shown on either beta-2 or beta-3.

MattSturgeon:
@BigAlanM, different issues should be discussed in separate threads. Instead of commenting on a random/recent issue/PR you should look for the most relevant one (or create it if one doesn't already exist).

In this case it sounds like you might be talking about #177 (which I found by searching for "village").

PR #392, opened 2018-04-23
Change Woodland Mansion icon to an evoker

MattSturgeon:
Inspired by @moulins' comment in #370 about the icon looking similar to the Nether Fortress icon, I thought I'd change it for an evoker icon:

image

I feel that using an evoker head represents the mansion well, providing enough contrast against other icons _(it is most similar to the stronghold icon in colour, but very different in shape)_. Additionally it compliments the existing village icon.

---

While at it, I also had a quick go at changing the witch hut icon from a cauldron to a witch head, however I ran into issues because the witch face is an even width while her hat is an odd width.


stefandollase:
Thanks for your contribution. It looks great =)

Issue #399, opened 2018-06-08
Implement changes to witch hut, jungle temple and igloo generation for 1.13


fwiffo:
Snapshot 18w06a made some changes to structure generation. In prior versions, desert temples, igloos, jungle temples and witch huts all shared the same magic values for seeding the RNG. Desert temples retain the same value, but the other three have new values in 1.13 and generate in different places than before.

See  also #395

fwiffo:
OK, after some digging I found a _bug_ in the new structure code for 1.13 which explains why structures aren't always appearing where I expect. It's going to reduce the likelihood of structures spawning in negative coordinates.

The correct way to find out which structure region a chunk belongs to requires special handling for negative numbers. E.g. for temples with a region size of 32:

```
offsetChunk = chunk < 0 ? chunk - 32 + 1 : chunk;
region = offsetChunk / 32;
```

That puts chunks -32 to -1 in region -1.

But the new code flipped that +1 to a -1. But then when converting back after selecting the random chunk for structure spawning, it does it the same old way.

```
chunk = region*32 + random(32-8);
```

So, let's say you're chunk -31, 0. You end up in region -2, 0 instead of -1, 0. If the RNG had picked 1, you would have been the lucky chunk and gotten an igloo. But instead, it thinks it should put the igloo in chunk -63. But chunk -63 is part of region -3. So if the RNG ever picks 0 or 1, it cannot generate a structure.

The other changes to the structure generation code are clearly deliberate, but this is unintended. I can implement it in amidst, but I'd hope they'd fix this. I've posted it on the minecraft bug-tracker [MC-131462](https://bugs.mojang.com/browse/MC-131462).

moulins:
Huh, that's interesting. So if I understand right, every Jungle Temple, Desert Temple, Igloo and Witch Hut with a chunk coordinate of the form `-32n` or `-32n + 1` won't generate?

If this bug is present since 18w06a, I think it should be implemented, even if Mojang fixes it. There already was a situation like this when Mojang increased the number of strongholds from 3 to 128 and the initial algorithm was buggy, and Amidst faithfully reproduces the bug (see [here](https://github.com/toolbox4minecraft/amidst/blob/master/src/main/java/amidst/mojangapi/world/icon/producer/StrongholdProducer_Buggy128Algorithm.java)).

fwiffo:
[...]

PR #400, opened 2018-06-09
Updates to structure generation

fwiffo:
For #399 - this patches cleanly against both the 1-13-snapshots branch as well as master.I went for a minimally invasive change, since it's really just a few constants that need to be updated. Adding three new options to the Layers menu didn't seem useful.


moulins:
This looks good, but isn't having `modifyNegativeCoordinate` as a `BiFunction` a little overkill? A simple boolean (enable or disable the bug) seems more appropriate.

Moreover, I think the way negative coordinates are handled (the `coordinate - distance ± 1`) is an implementation detail and should not be leaked outside the `StructureAlgorithm`.

fwiffo:
Yeah, you're right; changed it to a boolean. I just like writing lambdas because I can tell Java resents me for it.

fwiffo:
Thanks!

Issue #404, opened 2018-07-05
Some spawn point seems off target for versions since 18w06a


k-jiang:
It seems that some seeds have different spawn points (default home) in 1.13-pre5/6.

e.g. `-7358205204588637752`
In 1.12.2, the spawn point is (256,128), however in 1.13-pre6, the spawn point is (167,68) in game.

Not sure it is a intended game change or a snapshot bug.

moulins:
I tested your seed, and the first version for which the spawn point is inaccurate s `18w06a`. This coincide with the biome generation revamp, so it seems a (maybe unintended) consequence of that.

The version of Amidst which works for the latest snapshots still indicate the spawn point at (256,128), so the algorithm must have changed.

moulins:
**Note:** This may be related to #136.

PR #407, opened 2018-07-08
Support for 1.13 snapshots

### LINKED ISSUE #395: 1.13 snapshots not working
As you are probably aware, because of the refactorings made by Mojang, Amidst doesn't load correctly snapshots starting with `18w06a`.

I had a quick look this weekend, thinking I could relatively easily fix the issue, but there has been **lots** of change in the Minecraft code. After grabbing the latest version of the MCP mappings, and some help from the folks at the #mcp IRC, I have a rough idea of how the new worldgen code works.

-----
## Code changes in Minecraft
The worlgen code has been completely reworked, and this will affect Amidst in several ways:
 - Biome ids are now `int`s instead of `byte`s; this shouldn't be an issue for us as already use `short`s
 - The `ChunkProviderSettings` class has been completely reworked, and is now called `OverworldGenSettings`. I didn't find a way to construct an instance from an options `String`, and I suspect there isn't one for now, as the Customized world type doesn't exist anymore (there's the new Buffet type, but I don't have a clue how it works). We can still construct a default instance by calling its constructor though.
- Minor change, the `initializeAllBiomesGenerators` method has been moved to a different class called `LayerUtil`.
- The `IntCache` class is gone, this has to do with the next change (the most annoying one).
- The big change: the layer generation works differently now, and the `GenLayer::getInts` method is also gone. I'll go in more details, because this is important.

In previous versions, when generating biome data, each `GenLayer` would create an `int` array (grabbed from the `IntCache`) and populate it, and we could simply use the array returned by the final layers to access the biome data.  
Now, the `int` arrays are replaced by instances of the `IArea` interface, which has a single method:
```java 
public interface IArea {
   int getIntAt(int x, int y); //returns the data at the given coordinate
}
```
Each layer is an `IAreaFactory`, which acts as a transformer on these `IArea` objects.
**Note:** To avoid calculating several times data at the same coordinate, Mojang use a `LazyArea`, which cache recent computed values (I'm not entirely clear on the details).

The `GenLayer` class is now a small wrapper, whose job is to drive the computation and collect the result in the final array:
```java
public class GenLayer {
   private final IAreaFactory<LazyArea> factory;

   public GenLayer(IAreaFactory<LazyArea> factory) {
      this.factory = factory;
   }

   public Biome[] getBiomeData(int x, int y, int width, int height, @Nullable Biome defaultBiome) {
      AreaDimension areadimension = new AreaDimension(x, y, width, height);
      LazyArea lazyarea = this.factory.make(areadimension);
      Biome[] abiome = new Biome[width * height];

      for(int i = 0; i < height; ++i) {
         for(int j = 0; j < width; ++j) {
            abiome[j + i * width] = Biome.getBiomeFromId(lazyarea.getIntAt(j, i), defaultBiome);
         }
      }

      return abiome;
   }
}
```

## Impact on Amidst
If the code collected the data in an `int` array as before, this would be the end of it and the changes to Amidst would be minimal, but unfortunately the `getBiomeData` method returns a `Biome` array.

I see two ways to resolve this:
### Translating back into an `int` array
We can transform back the `Biome` object into its IDs, by using the Minecraft biome registry. To avoid calling into `SymbolicObject`s all the time, we should maintain a `Map<Biome, Integer>` on our side.

#### Pros
 - It should be straightforward to implement.

#### Cons
 - I'm wary of the performance implications of making a Map lookup for each data element. We may need a specialized Map type to have good performance here.
 - It forces us to do an allocation on each call (we can have a `int[]` in the `MinecraftInterface`, but the minecraft code will allocate a new `Biome[]` on each call).

### Using the underlying `LazyArea` directly
To avoid the theoretically useless roundtrip between `Biome` and `int`, we may be able to extract the `IAreaFactory` from the `GenLayer` and use it directly to collect the data into an `int` array. As before, we cannot use the `SymbolicObject` for every `getIntAt` call, so we need to somehow hoist this out the of the loop. The only way I see for now is to use a bytecode generator to generate the bytecode for the loop at runtime, but maybe there's a better way.

#### Pros
 - Better performance: no need for Map lookup, this should be as efficient as the minecraft code.
 - No allocation: we can reuse the same `int[]` to collect the data everytime.
 - It opens avenues to get even better improvements later, for example for the single-coordinates calls in structure checks.

#### Cons
 - It's complex to implement, and we need to dynamically generate bytecode to get acceptable performance.
 - We probably need an external library to do this, something like [ASM](http://asm.ow2.io/).

**Edit:** We can maybe get to close-to-native performance by using the MethodHandle API to call the Minecraft `LazyArea`, which should be *much* easier than writing bytecode!

## Phasing out the hardcoded biome list?

These changes may also be a good opportunity to remove the Amidst biome list, and instead use the Minecraft biome registry as a source of truth, as biomes now have an unambiguous id. This would be a good first step for eventually supporting modded Forge profiles with custom biomes in the long run.  
However, it's not that simple, because we still need a biome list somewhere to specify the colors we use in the map.

## Splitting the `MinecraftInterface` and the `ClassTranslator`

Whatever option we choose, I think it is also a good time to create a new `MinecraftInterface`, as proposed in #146. Even if the two versions (old and new) have some code in common, maintaining the two in the same classes will quickly become a nightmare.  
I've pushed [a branch](https://github.com/toolbox4minecraft/amidst/tree/1-13-snapshots) with some quick-on-dirty modifications to dispatch the correct `MinecraftInterface` based on the version, with a stub for the `1.13` implementation, go criticize it! (This doesn't use `VersionFeatures` yet, as I didn't see a good way to use it here)

------

Phew, that was longer than expected!
Feel free to give your thoughts on this.

moulins:
This PR adds supports for 18w06a and newer versions.
 - Add a new `MinecraftInterface` for 18w06a and later;
 - Add 1.13 Ocean biomes;
 - Add Ocean Ruins and Shipwrecks;
 - Remove M biomes which don't exist in the Minecraft code;
 - Add the `benchmarkWorldGeneration` dev tool, to measure the performance of the world generation.

There are still some issues:
 - Biome generation is veeery slow (~ 30x slower than 1.12), but sadly, we can't do much about it.  (Maybe generating biome data in parallel could help?)
 - Some seeds have seen their spawn point changed in 18w06a; this PR doesn't account for that (see #404).

Even then, I think this is a good time to merge this, as Minecraft 1.13 will soon be out.

Fixes #395.


fwiffo:
Regarding biome generation speed... When I was last digging into the code, it was not thread-safe, unless they've fixed it quite recently. It uses this cache class which is full of statics. So fixing that, without modifying the Minecraft code itself, would be pretty hard (but I don't know Java, so maybe there is some crazy class injection thing that can be done).

There is also [Cubitect's cubiomes](https://github.com/Cubitect/cubiomes) C reimplementation of Minecraft biome generation, which is much, much faster straight out of the box, and is also thread-safe, so it can be made faster still (I've been using it for seed-finder code, and it works splendidly.) Again, I don't know how practical that is to use in a Java project, and it's also got to be kept up to date with new Minecraft versions automatically. It's only partially updated to 1.13.

moulins:
> It uses this cache class which is full of statics.

If you're talking about the `IntCache` class, it's gone in 18w06a (in fact, this was one of the things which broke Amidst). So, unless there are other shenanigans I haven't seen, the biome generation should be thread-safe now.

> There is also Cubitect's cubiomes C reimplementation of Minecraft biome generation ...

Well, apart from the fact that using C code isn't really possible (it's a pain in the *ss to use C code from the JVM, and the overhead is significant), one of the strengths of Amidst is that it uses Minecraft's own code, so it can always provide accurate biome data. Using a third-party reimplementation of the biome generation would defeat the purpose.

fwiffo:
Yay, death to IntCache! From my testing, that's the only thread-safety issue.

Regarding the C implementation: Yeah, that's what I figured. Oh well.

seaside98:
It doesn't look like you have added the new ocean structures to the layers dropdown in the menu yet. Would be nice to have those there.

kylerchin:
+1

seaside98:
You could use 0-9 in addition to ctrl+0-9.

moulins:
I did finally separate the new Ocean structures into a new layer, which is toggled by `ctrl+0`. It is the same shortcut as the End Cities layer, but because they are in different dimensions one of the two will always be disabled, and thus the overlap doesn't cause issues.

PR #420, opened 2018-08-04
MC-131462 bug was actually fixed in 18w30a.

fwiffo:


moulins:
Amidst policy regarding is to only record the last snapshot of the week (so for example, if there is a `18w30b`, `18w30a` won't be present in the list), to limit the size of the recognised version list. This is also what is done by the Minecraft launcher.

As such, we use `18w30b` for the version of the bugfix, instead of `18w30a`.

I'll be closing this.

PR #498, opened 2019-06-08
Move JNBT library to local repository

CubeTheThird:
Instead of installing the jnbt dependency during the clean phase, it can be instead referenced directly from a local repository within the project.
This can circumvent issues where it is not installed when clean is not called, or when combined with other phases.


moulins:
Looks good, thanks!

PR #499, opened 2019-06-10
change biome names to reflect 1.14

nstickney:
Not sure if there was a reason the names haven't been updated yet, but I thought I'd throw this pull request in just in case it was because no one had time to update them yet. If it's not helpful, no worries!


MattSturgeon:
The new names should probably only be used for minecraft versions affected by the change. Which snapshot did they do this in again?

nstickney:
1.13pre5, according to the wiki.

moulins:
This is something that I want to do, but as @MattSturgeon said, we need to keep the old biomes names in previous versions of Minecraft.

The biome list is currently static and can't change between profiles, so this seems hard to do with the current architecture.

I'll close this PR for now, but this is definitely something which should be done once the biome list is made dynamic.

burgerguy:
This really shouldn’t be too difficult. We have many different options to go about doing this.

Option 1: We retrieve the biome names from the minecraft registry in a loop upon opening a profile.

Option 2: We put the biome names in the version features and manually enter each time the name has changed.

Option 1 would be better because we wouldn’t have to update them in the future, but we have to rely on the MinecraftInterface and we have to keep track of if the biome names have been retrieved or not when using them for the UI. Option 2, however, would be easier but would take more time and be more of a hassle to keep updated.

moulins:
@burgerguy I have recently been working on a branch to allow more flexible `VersionFeature`s (see PR #696). With this change, we could have the list of `Biome` as a version feature, and have a mix of your options 1 and 2:
- Add a `Optional<List<Biome>> getBiomeList()` to `MinecraftInterface`, which tries to get the biome list from the loaded Minecraft instance;
- If this is unsuccessful (for example, for old Minecraft versions), fall back to the current hardcoded list.

PR #542, opened 2019-10-07
Use openjdk instead of oraclejdk for Travis CI.

heuermh:
`oraclejdk8` is no longer supported by Travis CI.


moulins:
Thanks for the fix!

heuermh:
Thank you, @moulins!

Issue #538, opened 2019-09-28
Snapshots 19w36a+ aren't supported


Luneia:
[info] Amidst v4.3
[info] Current system time: 2019-09-28 10:27:07.73
[info] System.getProperty("os.name") == 'Windows 10'
[info] System.getProperty("os.version") == '10.0'
[info] System.getProperty("os.arch") == 'amd64'
[info] System.getProperty("java.version") == '1.8.0_221'
[info] System.getProperty("java.vendor") == 'Oracle Corporation'
[info] System.getProperty("sun.arch.data.model") == '64'
[info] Not using OpenGL.
[info] Biome profile activated: default
[info] using '.minecraft' directory at: 'C:\Users\nomis\AppData\Roaming\.minecraft'
[info] using biome profiles at: 'biome'
[info] Starting to download remote version list.
[info] Scanning for profiles.
[info] Successfully loaded profile list.
[info] There are no updates available.
[info] Successfully loaded remote version list.
[info] using minecraft launcher profile 'Latest snapshot' with versionId '19w39a'
[info] Loading libraries.
[info] Found library com.mojang:patchy:1.1 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/com/mojang/patchy/1.1/patchy-1.1.jar
[info] Found library oshi-project:oshi-core:1.1 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/oshi-project/oshi-core/1.1/oshi-core-1.1.jar
[info] Found library net.java.dev.jna:jna:4.4.0 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/net/java/dev/jna/jna/4.4.0/jna-4.4.0.jar
[info] Found library net.java.dev.jna:platform:3.4.0 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/net/java/dev/jna/platform/3.4.0/platform-3.4.0.jar
[info] Found library com.ibm.icu:icu4j-core-mojang:51.2 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/com/ibm/icu/icu4j-core-mojang/51.2/icu4j-core-mojang-51.2.jar
[info] Found library com.mojang:javabridge:1.0.22 at file:/C:/Users/nomis/AppData/Roaming/.minecraft/libraries/com/mojang/javabridge/1.0.22/javabridge-1.0.22.jar
[...]

PR #545, opened 2019-10-10
[WIP] Experimental support for 1.15 snapshots

moulins:
See issue #538.

In `19w36a`, the Minecraft code for biome generation was updated, with two changes affecting Amidst:
- The `GenLayer.getBiomeData(x, y, width, height)` was removed;
- The full-resolution biome data is no longer available as a `GenLayer`: it is now managed by a separate `BiomeZoomer` interface.

As such, we need a different approach to access the biome data, and a new `MinecraftInterface` implementation.

This PR add an experimental implementation, which works on versions `19w36a` through `19w41a` (the current version). The performance is roughly comparable with the current `1.14` implementation.

I'll wait some time before merging this (maybe until the 1.15 release?), to not have needless churn if the Minecraft code change again.


burgerguy:
The optimizations for 1.14 that I have made should be compatible with 1.15 as it does not rely on the `GenLayer.getBiomeData` function

odegroot:
Minecraft-1.15 is out now. 😃

PR #555, opened 2019-10-20
Add -seed command line argument

magicus:
Add -seed command line argument to start Amidst with a particular world seed selected. This implements feature #461.

I've tried my best to follow https://github.com/toolbox4minecraft/amidst/wiki/Supporting-the-Development. I have announced my intention in #461, unfortunately I have not received any response from the maintainers. :-( If this PR is not accepted, please let me know what I need to fix rather than just close it outright.

### DIFF
--- a/src/main/java/amidst/CommandLineParameters.java
+++ b/src/main/java/amidst/CommandLineParameters.java
@@ -28,7 +28,10 @@ public class CommandLineParameters {
 
 	@Option(name = "-log",                    usage = "location of the log file",                            metaVar = "<file>")
 	public volatile String logFile;
-	
+
+	@Option(name = "-seed",                   usage = "initial seed to use",                                         metaVar = "<string>")
+	public volatile String initialSeed;
+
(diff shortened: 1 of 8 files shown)


moulins:
This looks nice, but I don't like the fact that the world type is hard-coded to `DEFAULT`.

It would be best to add optional CLI parameters to specify the world type and the generator options (`-world-type` and `-world-options`, for example), and replace all the `String initialSeed` parameters by `WorldOptions initialOptions`.

magicus:
I've added a -world-type command line option. This will only function in tandem with -seed. (It made no sense to specify an initial world type but no seed; it's more likely that the user made an error rather than wanting to pre-specify a world type for the next GUI interaction.)

The additional generatorOptions in WorldOptions is currently not used, so I did not give them a separate command line option. (For starters, I did not even know what they could be since I had no examples to work on.)

moulins:
Thanks!

PR #560, opened 2019-10-26
Add -profile as a CLI option for easier selection of profile.

magicus:
This is the second part of #461. This patch adds a -profile command line option. While technically equivalent to using -mcjson and -mcjar, this makes for a simpler user interface. You can just specify "-profile 1.32.2" (if that's what your profile is called) to select it, instead of looking up the correct path to your minecraft directory and the corresponding version json and jar files.

-profile cannot be used together with -mcjson and -mcjar. I'm limiting the use to local profiles (I don't fully understand the remote profiles, and I don't want to rely on remote query to properly parse command line options), but as far as I understand, this is the same limitation that using -mcjar and -mcjson sets.

I also added a patch for .gitignore. It was bugging me that it didn't work correctly with IntelliJ IDEA.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,9 @@
 *.ear
 
 # IntelliJ IDEA project files #
-./AMIDST.eml
-./AMIDST.iml
-./.idea/*
+/AMIDST.eml
+/AMIDST.iml
+/.idea/*
 /bin
 /dist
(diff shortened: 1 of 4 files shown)


moulins:
Looks good to me, thanks!

Issue #451, opened 2018-09-10
[feature reguest] cache map when scrolling


DmitryRendov:
I understand that it might be difficult to implement, but there's no need to render already rendered chunks)
thanks!

lams2135:
it would be a really awesome improvement if implemented. we need that.

josephrooks:
@moulins I'd like to add to this, if there's a way to stop it from rendering biomes at all, would that speed up its ability to find coordinates? I don't need the visuals at all if I'm using it to look for woodland mansions or triple/quad witch huts, for example.

moulins:
@josephrooks You would probably get a small speed up, but Amidst still needs to inspect the biome data in order to determine is a structure location is valid, so the speed increase may not be very big.

If you don't need any visualisation and only want raw speed, there are more specialized tools, like [this quad hut finder](https://github.com/fwiffo/SciCraftSeedFinder).

magicus:
It turns out this was not very hard to do. I have a proof of concept with a simple WeakHashMap cache in FragmentManager.requestFragment(). The resulting change in responsiveness in Amidst is impressive, almost mind-blowing! If you zoom in and then back out, everything's just cached and displayed immediately. Same thing is you scroll back and forth in the map. Instant re-display.

So this is definitely a path worth pursuing. Of course, the devil is in the details. Finding the right limits of the cache, a good retention policy, etc, will take some more work. Also, Amidst is currently very eager to invalidate and re-use allocated Fragment objects in an attempt to pick a fight with the garbage collector. (This is generally not a very good idea.) To make the cache work, I just temporarily commented out FragmentQueueProcessor.recycleFragment(). For a patch that would be acceptable, I need to find a better solution. The start of that would mean that I need to understand the Fragment lifecycle a bit better.

magicus:
I'm getting closer to an acceptable patch. The problem I'm running into is that the entire design around how fragments are allocated, calculated ("loaded") and finally removed is ... eh, non-optimal. There's a lot of stuff going on that's not very efficient, or good coding practice. I can add in a caching mechanism on top of this, with a minimal patch, but I think it would be beneficial to restructure the code, to simplify the logic and lifecycle of Fragments.

@moulins Would your be prepared to accept a somewhat larger patch that straightens out the Fragment handling? I think it will be beneficial for the code base. Otherwise I'll go ahead and try to make the best of the current situation.

magicus:
[...]

PR #564, opened 2019-10-29
Cache Fragments when off-screen

magicus:
This PR implements caching of Fragments, as requested and discussed in #451. 

The SoftRef implementation guarantees that the GC will free elements in the cache before running out of memory. The cache will be invalidated whenever the complete set of fragments is to be recycled, i.e. when we're switching seed, etc. Loading of new fragments will proceed exactly as before, unless the fragment is found in the cache, in which case it will be reused. A fragment that was requested to be loaded at some point (e.g. by the user scrolling to that part), but then was retired (by that part going off the screen again) will still be eligible for loading, but in the background, when all visible fragments has been loaded. That way, if a user scrolls this way and that, chances are the fragment is loaded and cached when she returns to that part of the map.

### DIFF
--- a/src/main/java/amidst/CommandLineParameters.java
+++ b/src/main/java/amidst/CommandLineParameters.java
@@ -28,7 +28,10 @@ public class CommandLineParameters {
 
 	@Option(name = "-log",                    usage = "location of the log file",                            metaVar = "<file>")
 	public volatile String logFile;
-	
+
+	@Option(name = "-seed",                   usage = "initial seed to use",                                         metaVar = "<string>")
+	public volatile String initialSeed;
+
 	@Option(name = "-help",                   usage = "print usage information")
 	public volatile boolean printHelp;
 
--- a/src/main/java/amidst/CommandLineParameters.java
+++ b/src/main/java/amidst/CommandLineParameters.java
@@ -29,8 +29,11 @@ public class CommandLineParameters {
 	@Option(name = "-log",                    usage = "location of the log file",                            metaVar = "<file>")
 	public volatile String logFile;
 
-	@Option(name = "-seed",                   usage = "initial seed to use",                                         metaVar = "<string>")
+	@Option(name = "-seed",                   usage = "initial seed to use",                                 metaVar = "<string>")
 	public volatile String initialSeed;
+
+	@Option(name = "-world-type",             usage = "world type for the initial seed",                     metaVar = "<string>")
+	public volatile String initialWorldType;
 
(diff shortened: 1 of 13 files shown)


burgerguy:
Amidst not doing this currently is kind of a large problem. If you want to zoom in on an area for a short amount of time, when you zoom out, you have to wait for it to regenerate all of the fragments again.

magicus:
@moulins Is there anything holding back this PR being merged? If you need any specific changes, please let me know. I'll be happy to update it to fit with the projects style guide, as well as long-term vision.

moulins:
Sorry, I didn't have time to test the PR myself.

-----

I did have time to run some tests today, and while moving the map around for a few minutes, RAM usage kept increasing. It seems that the background fragments never get garbage-collected, and when I stopped moving around the fragment cache has grown to more than 10k entries.

I see two potential causes for this:
- when a fragment is retired, it is added to the background queue, which means it can't be reclaimed by the GC until it is fully loaded by the fragment loader;
- in the `FragmentManager`, there is both `FragmentCache cache` and `Map<CoordinatesInWorld, SoftReference<Fragment>> fragmentCache`. I'm not entirely sure this is a problem, but I think it would be clearer if the `Map` was inside the `FragmentCache` instead.

In addition, it would be nice (and help with debugging) if the status of the background queue was displayed by the `DebugWidget` alongside the other queues.

burgerguy:
I've been recently working on a branch ([here](https://github.com/burgerguy/amidst-changes/tree/multithreaded-fragment-cache)) which rewrites most of the fragment queues and caches fragments. In short, the normal `FragmentCache` was removed, as well as the `availableQueue`. Now, there are separate classes `OffScreenFragmentCache` and `AvailableFragmentCache`. The off-screen fragments have a limited lifespan of 30 seconds so they don't take up too much ram, and they are also `SoftReference`s so they don't cause an OOME. The `AvailableFragmentCache` is in place of the old `availableQueue` and adds some extra features. It uses `SoftReference`s, but they don't have an expiration time. A thread gets spawned to clean and maintain both of these caches every half a second.

With the `FragmentCache` being gone, this means that no cache will have the on-screen fragments once they are all done loading. We now have to get them from the `FragmentGraph` instead, and it's used for when we need to invalidate a layer and put the fragments back in the `loadingQueue`.

Overall, it should make the fragment pipeline a bit simpler and should also use a lot less ram.

magicus:
Nevermind this PR then...

I've frankly given up on Amidtst myself anyway; the chunkbase javascript app is way faster. (I think this state of affairs is a bit sad, though.)

PR #589, opened 2019-12-18
Rewrite the current LocalMinecraftInterface for performance gains

burgerguy:
Instead of using BiomeZoomer, we can use LazyArea and then zoom it ourselves. This allows us to use previous performance improvements that were used with LazyArea. I also changed some RealClass stuff to detect classes more reliably and easily.

### DIFF
--- a/src/main/java/amidst/clazz/real/RealClass.java
+++ b/src/main/java/amidst/clazz/real/RealClass.java
@@ -1,6 +1,7 @@
 package amidst.clazz.real;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -46,6 +47,7 @@ public class RealClass {
 	private final List<Float> floatConstants;
 	private final List<Double> doubleConstants;
 	private final List<Long> longConstants;
+	private final List<Integer> intConstants;
 	private final List<Integer> stringIndices;
 	private final List<ReferenceIndex> methodIndices;
 
@@ -68,6 +70,7 @@ public class RealClass {
 			List<Float> floatConstants,
 			List<Double> doubleConstants,
 			List<Long> longConstants,
+			List<Integer> intConstants,
 			List<Integer> stringIndices,
 			List<ReferenceIndex> methodIndices,
 			int accessFlags,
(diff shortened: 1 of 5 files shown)


moulins:
Thanks for the PR!

I've done some timing tests comparing the current `master` code and your code, including some variations (`master` is `Biome` access + simple loop; your PR is `int` access + chunked loop). Here are the results for rendering a fully-dezoomed map:

|   | `Biome` access | `int` access |
|----|----|----|
| **Simple loop** | 43s | 42s |
| **Chunked loop** | 23s | 22s |

The direct `int` access *via* the `PixelTransformer` doesn't seem necessary to get the performance improvements of the chunked loop. Given that using the `PixelTransformer` forces us to reimplement the `BiomeZoomer` ourselves, I won't merge this PR, and will instead make a commit introducing the chunked loop into the current code.

burgerguy:
The reason for getting the data from the PixelTransformer is so that we bypass the synchronized block in the LazyArea get() method shown here:

```java
   public int get(int var1, int var2) {
      long var3 = ChunkPos.asLong(var1, var2);
      synchronized(this.cache) {     // here
         int var6 = this.cache.get(var3);
         if (var6 != Integer.MIN_VALUE) {
            return var6;
         } else {
            int var7 = this.transformer.apply(var1, var2);
            this.cache.put(var3, var7);
            if (this.cache.size() > this.maxCache) {
               for(int var8 = 0; var8 < this.maxCache / 16; ++var8) {
                  this.cache.removeFirstInt();
               }
            }

            return var7;
         }
      }
   }
```

The reason the synchronized block exists is so that it can interface with it's (the LazyArea's) cache safely. Because we don't need the cache, we don't need the synchronized. This eliminates the possibility of a bottleneck for multithreading once implemented. Currently in the state of single threaded biome data collection, this doesn't really benefit a whole lot, but once multithreading is implemented, this could help performance a lot. This can also be translated back to 1.13/1.14 so there wouldn't be a need for a seperate minecraft interface.

moulins:
For future multithreading, wouldn't it be enough to create a separate `BiomeProvider` for each thread? (I think this is what Minecraft does when generating new chunks).

burgerguy:
I’d need to look into the minecraft code to see what sort of strategy they use for multithreading in that case. It probably ends up resulting in calling LazyArea.make() for every time it wants to make a new BiomeProvider, which would be unnecessary and cause more overhead. On the other hand, it may be easier to debug when doing it the same way as minecraft, and we may be able to do all of the multithreading through minecraft without making our own threads.

burgerguy:
I couldn’t figure out where minecraft starts multithreading, so instead I fixed the full resolution biome data and made a new pull request (#649).

PR #649, opened 2020-01-13
Changes to LocalMinecraftInterface for later multithreading, etc.

burgerguy:
Changes include:
* Fixed up some things in RealClass and added a search for int constants
* Completely changed LocalMinecraftInterface to use PixelTransformer, with a proxy interface and int to biome to int conversion for getting the full resolution data
* ~Changed the method used for getting single block full resolution biome data in BiomeDataOracle (yes I know it's ugly but I didn't want to remove it)~ Removed

I looked for around 20 minutes trying to find the biome multithreading code to hook on to in minecraft, but I couldn't find anything. So instead, I fixed the workaround of implementing the minecraft code directly. Annoyingly, it ended up requiring me to use methods for conversion between `int` and `Biome` again. Fortunately, it was only about 30%(Updated, see edit) slower than the direct method, and calling for full resolution biome data doesn't happen often. The proxy stuff that I use is kind of janky, but it is surprisingly fast and gets the job done. I'm committing this before I start with any multithreading code so I don't get any further behind if more code is changed.

This is an addition to an earlier pull request (#589) which instead includes a portion of the minecraft code implemented into the LocalMinecraftInterface.

Edit: after re-profiling, I realized that this method surprisingly only dropped about 10% performance in full resolution getting.

Edit 2: typo.


burgerguy:
I did some experimenting with MethodHandles, and it seems that they are about as slow, if not slower, than normal reflection. We could try to use LambdaMetafactory, but it looks really complicated and there isn't a whole lot of documentation on it.

burgerguy:
I found out that when skipping the first LazyArea to get the PixelTransformer, there are still many other LazyAreas further down the line that end up synchronizing anyway. I tried doing a proxy based solution to get rid of LazyAreas entirely, but it ended up being extremely slow due to how proxys work. At some point we could possibly use a bytecode manipulation library to change the LazyAreaLayerContext to create uncached areas of some sort. For now, however, I'm removing all PixelTransformer code and reverting to LazyAreas.

burgerguy:
Going to close this pull and make a new pull due to a mostly successful multithreading implementation

PR #652, opened 2020-01-14
First commit of Look and Feel changes

burgerguy:
If someone can test this on macOS that would be great, because currently I'm not able to test it there.

### DIFF
--- a/src/main/java/amidst/Amidst.java
+++ b/src/main/java/amidst/Amidst.java
@@ -5,7 +5,9 @@ import java.sql.Timestamp;
 import java.util.Date;
 import java.util.prefs.Preferences;
 
+import javax.swing.JFrame;
 import javax.swing.SwingUtilities;
+import javax.swing.UIManager;
 
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
@@ -79,6 +81,7 @@ public class Amidst {
 			AmidstLogger.info(versionString);
 			logTimeAndProperties();
 			enableGraphicsAcceleration();
+			setLookAndFeel();
 			osxMenuProperties();
 			startApplication(parameters, metadata, createSettings());
 		}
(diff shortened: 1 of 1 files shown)


EdricChan03:
Testing locally on my end, it seems that the look and feel isn't changed on macOS. I think it's probably because the look and feel class you're using doesn't exist on macOS.

I think it would be best to use the [`UIManager#getSystemLookAndFeelClassName()`](https://docs.oracle.com/javase/8/docs/api/javax/swing/UIManager.html#getSystemLookAndFeelClassName--) method instead of the `NimbusLookAndFeel` class.

moulins:
On my machine, I can use `UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName())` to use the system look & feel; it seems it should work on OS X.

However, I would like the user to be able to select which look & feel to use between the default L&F and the system L&F, in case the system L&F breaks something in the UI (maybe with a checkbox in the "Settings" menu?). I'll look into it in the next few days.

burgerguy:
I made the Nimbus look and feel was not available due to it crashing on startup. The same LaF code should work on macOS considering I had an extremely similar setup for another program and I basically just copy-pasted the code from there, and in that case I had a chance to test it on macOS. I am not using the NimbusLookAndFeel class here. I am checking if it is what is naturally selected by the system to be able to revert back to the normal one.

If you are looking on having a checkbox to change the look and feel, please remember that not all operating systems support eachother's look and feels. Most of them use natives that are build into their OS's, with the exception of Nimbus and the default. Also, after changing the look and feel, remember to call `SwingUtilities.updateComponentTreeUI(jframe)` to allow the components to update. 

This is probably not a good idea, however, and could possibly break some things. A better option is to have a setting that only applies after a restart of the program, and have a prompt that shows up if it crashes initializing the look and feel to change that setting to the default.

moulins:
This is superseded by #662; closing.

Issue #599, opened 2019-12-23
End generation fails at far distances from end spawn


ghost:
the used operating system and version - Windows 10 x64 Version 1903

the used Java version - Version 8 Update 231 (build 1.8.0_231-b11)

the used Minecraft version - 1.12.2

the used Amidst version 4.3

the error message, if any - no error message

What did you do to produce the error? - typed in the coordinates to my end base far from spawn and noticed the end not generating correctly that far out.

(How) can the error be reproduced? - Change the map view to the end, then go to the coordinates       (0, 1,048,576). It fails at similar distances on the other axis. Also, when inputting coordinates close to or at the world border (30,000,000 , 30,000,000) the end seemed to generate in a striped pattern.

What did you expect to happen? - Amidst should still show end terrain.

What did actually happen? - Amidst generates no terrain beyond that point.

Examples of glitched generation:
    No Generation - https://imgur.com/a/YbiUYVv
    Striped Generation - https://imgur.com/a/n2NwJ2r

burgerguy:
Probably the weirdest thing I've seen with amidst. I'll look into it, but it looks like that number is 2<sup>20</sup>.

burgerguy:
It was caused by an integer overflow in two different places. In one, I just caught if there was an overflow, and in the other I instead did the calculations with longs.

James103:
I assume that this is related to https://www.youtube.com/watch?v=91Feq0dHw28 because some similar overflows took place. In fact, both 741455 and 1048576 are x-intercepts of the function `y = sin(x^2 * 0.0000000000228580929)`, which is the same function mentioned in the above video (that explains why the End terrain generation was broken in the first place).

moulins:
Huh, that's interesting.  

Maybe we should reimplement the bug "correctly" for the affected Minecraft version?

James103:
Maybe add an overlay that makes the End terrain partially transparent if the function `f(x) = sin(z^2 * π / 137438953472)` is negative, where `z^2 = x^2 + y^2` (the distance from 0,0 to the coordinate being displayed)?

burgerguy:
This would actually be really easy to implement, so I might do this when I get the chance

PR #653, opened 2020-01-15
Fixes bug #599 caused by integer over/underflow

burgerguy:


moulins:
I already had a fix implemented locally (2d0865ed) so I'll close this; but thanks anyway!

Issue #650, opened 2020-01-13
Menu item naming inconsistencies (macOS)


EdricChan03:
Currently, the application has some menu item naming inconsistencies which are shown in the screenshots below:

![Screenshot of "World" menu](https://user-images.githubusercontent.com/20047125/72276738-9eb0aa00-366b-11ea-9788-a3d0872b7a1a.jpg)

A couple of issues with the menu above:

- Inconsistent capitalisation (e.g. "Save player locations" vs "Take Screenshot...")
- Menu item for a specific question (this could be moved to a help button inside of the functionality itself)

![Screenshot of "File" menu](https://user-images.githubusercontent.com/20047125/72276800-b851f180-366b-11ea-9c6b-43bd77dd4dc1.jpg)

More issues with the menu above:

- Same as before.
- There's no need for an "Exit" menu item on macOS when macOS already provides a "Quit [app]" menu item by default in the application's menu.
- Ellipses should be appropriately added to menu items following the suggestion below:
  > **Use an ellipsis whenever choosing a menu item requires additional input from the user.** The ellipsis character (…) means a dialog or separate window will open and prompt the user for additional information or to make a choice. - [Human Interface Guidelines for Menu Anatomy > Menu Item Titles](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/#menu-item-titles)

As documented in the [Human Interface Guidelines for Menu Anatomy](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/#menu-and-menu-item-titles), it suggests the following regarding naming menus/menu items:

> **Use title-style capitalization.** Title-style capitalization is used consistently for menu and menu item titles throughout the system. For more information, see [Apple Style Guide](https://help.apple.com/applestyleguide/#/apsg1eef9171).

Additionally, in v4.4-beta1 of Amidst (which introduced support for a native menu bar for macOS - see #540), the "Layers" menu does not seem to do anything until you enter a seed/random seed/open a save game. As such, I suggest that the menu should be hidden unless a seed is entered.

---

EDIT: I've forgotten about the "Settings" menu - I think it should be moved to the application menu under a menu item named "Preferences..." with a keyboard shortcut of <kbd>Command</kbd> + <kbd>;</kbd> like what most other native macOS programs use. A dialog would then show up with a set of checkboxes similar to the existing settings menu items.

burgerguy:
Addressed some of this in PR #651

[...]

PR #662, opened 2020-01-18
Add setting to select the look and feel

moulins:
Add the `Settings > Select Look and Feel` menu item, allowing the user to choose whether to use the cross-platform L&F (default choice) or the L&F provided by the OS.


EdricChan03:
Testing the PR locally, it seems that setting the look and feel to "System" has no effect on the default font. Here are some screenshots of the changes:

Screenshot of confirmation look and feel dialog

Additionally, could the "Settings" menu be moved to a dialog in the future? It can be troublesome to access settings in a menu. (I suggest to move this to a "Preferences..." menu item in the app's menu) This could be done in a future PR.

moulins:
> Additionally, could the "Settings" menu be moved to a dialog in the future? It can be troublesome to access settings in a menu. (I suggest to move this to a "Preferences..." menu item in the app's menu) This could be done in a future PR.

As I've said in #650, I'm not even sure this is a good idea, so I prefer to not do it yet.

PR #693, opened 2020-02-09
Add 20w06a to versions list, fix ocean features placement for 20w06a

burgerguy:
Also fixes some seeds being impossible to generate when making a random seed


moulins:
Thanks for the changes!
I added a commit to use a static `SecureRandom` instance for the seed generation, because initializing a new `SecureRandom` for each `WorldSeed.random` call is needlessly expensive.

PR #698, opened 2020-02-14
After loading registry, stop datafixer threads

burgerguy:
The threads that register all of the datafixers take up a TON of CPU usage for a few minutes after loading the Registry class, and it doesn't help us in any way. We can just stop them a bit after they are created by getting the ExecutorService running them. I couldn't find a way to load the class without the threads being created, so I just did this instead.

Also instead of manually importing every field in the SymbolicNames, I just imported all of them. I feel like there may have been a reason for this, so feel free to change it back.

**EDIT**: The reason I don't use the method for this built into `Util` is because it requests a shutdown to the threads, waits a few seconds for them to shutdown, and then if it isn't already shut down, forcefully shuts the threads down. In my testing, the threads always had to be forcefully shut down. The way I implemented it skips the few seconds added to loading time.

### DIFF
--- a/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java
+++ b/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java
@@ -97,7 +97,8 @@ public enum DefaultClassTranslator {
                 .thenDeclareRequired(CLASS_BIOME)
             .next()
 				.ifDetect(c -> 
-					c.searchForStringContaining("Server-Worker-")
+					(c.searchForStringContaining("Server-Worker-")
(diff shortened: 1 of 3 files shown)


moulins:
Nice work! I had already noticed this CPU spike, but I had chalked it up to the class loading itself. :stuck_out_tongue:

However, it doesn't seem to work in 20w06a (I get the `[warning] Unable to shut down Server-Worker threads` message).

burgerguy:
Sorry for the late reply, I'll take a look to see what's causing this.

burgerguy:
Looks like they changed the worker thread names from "Server-Worker-" to "Worker-". I use this term to search for the class, which is why it wasn't able to be found in 20w06a. Overall a pretty simple fix.

PR #700, opened 2020-02-23
Multithreaded fragment loading and 1.13/LocalMinecraftInterface rewrite

burgerguy:
(Built off #589 and #649. If you want to know more about my LocalMinecraftInterface implementation, you should probably start there.)

This barely works in its current state, but loads fragments extremely fast. It took about 4 seconds to load a fully zoomed out window with 12 threads, but it leaves a couple holes and messed up fragments. Strangely, if you hover over the holes and messed up fragments, it shows the correct biomes in the top right.

Initially I was hesitant to make multiple `PixelTransformer`s at once without cloning. In my testing, however, it was much faster to create one than I thought, with a single transformer taking about 0.06 seconds to create.

The reason I'm still using `PixelTransformer`s instead of `BiomeProvider`s is because it's slightly faster, bypasses a lot of unnecessary constructors for things like WorldType, GameType, and WorldData, and avoids some extra stuff with the registry for the `BiomeProvider`s themselves. Another factor is that i'm obviously much more familiar with my implementation because I've been working with it a lot more. If you want to me to port it to work with `BiomeProvider` for whatever reason then feel free to let me know.

I'll create this pull as a draft for now, but when I feel as if the implementation is stable enough, I'll make it a normal pull.


burgerguy:
I realized that this completely screws up the end currently, so I'll have to find a way to keep only the overworld biome data collection multithreaded.

burgerguy:
After thinking about the issue of the messed up fragments for a bit, I now realize that this is a much deeper problem than I anticipated. For one, I realize that this happening due to the rendering system relying on the loaders loading in order, which isn't always going to happen in situations like this. This comes back to the fact that to draw things on top of each other with `Graphics2D`, you need to draw the things in order. This won't work when we're using different `FragmentLoader`s at different times. To get around this, we need to have two stages of loading: stage one is the processing of everything that is needed, and stage two is actually drawing it on the screen. We also need to make sure that the `ImageLoader` is **always** loaded after the `BiomeDataLoader` due to the nature of how the ImageLoader works. There may be other ways to go about this, but this should be the easiest way to do it in the current framework.

burgerguy:
Making the `ImageLoader` thread safe fixed about 97% of the broken fragments, while retaing the same speed. This PR is already close to done, and in my opinion this should also be merged in combination with @magicus's pull for caching previously rendered fragments (#564). I was able to combine the two together very easily, and is available [here](https://github.com/burgerguy/amidst-changes/tree/magicus-magicus-fragment-caching) for testing.

burgerguy:
It seems like the bug in question does not occur (need to verify) on magicus's fragment cache system, but the bug seems quite difficult to reproduce. The easiest way I've found is to zoom out really quickly or move the screen really quickly, but that doesn't seem to always work. When it happens though, it shows up out of order of rendering, and is always incorrect. The actual rendering process doesn't seem to try to correct it when passing over those fragments. The fragments circled in the image are the ones with the bug in question. I may have missed a few, but you get the idea. This seems to be the only bug left, and the next thing to do is to port this implementation to 1.13-1.14 as it shouldn't be too hard.

amidst

burgerguy:
I haven't fixed that bug yet, but I have added a slider for the amount of threads used in fragment loading as well as an updated `_1_13_MinecraftInterface` which is much, much faster than it was previously. When I ran it with 1.14.4, I got results comparable to 1.12.2 speeds. 1.13.2 speeds were comparable to the 1.15.2 speeds. I'm looking forward to fixing the bugs left in the implementation, because this is a needed change for things like seed search to be a possibility again.

burgerguy:
I have whittled the possibilities of the bug down to something with the `BufferedImage`s.

moulins:
Sorry for the delay; this is a big change, so it's taken me time to review it.

----- 

> The reason I'm still using PixelTransformers instead of BiomeProviders is because it's slightly faster, bypasses a lot of unnecessary constructors for things like WorldType, GameType, and WorldData, and avoids some extra stuff with the registry for the BiomeProviders themselves. Another factor is that i'm obviously much more familiar with my implementation because I've been working with it a lot more. If you want to me to port it to work with BiomeProvider for whatever reason then feel free to let me know.

I do think it would be beneficial to reimplement the threaded changes with `BiomeProvider`s, for several reasons:
- The current PR mixes the `PixelTransformer` changes with the changes needed for multi-threading `MinecraftInterface`s; I think they would better be separate.
- Similarly, removing the `PixelTransformer` changes would allow to better measure the performance gains of the multi-threading. (Additionally, when I last tested your version of `LocalMinecraftInterface`, I didn't see any performance improvements, so I'm not even sure these changes are beneficial).
- I'm not fond of the [change in BiomeDataOracle](https://github.com/toolbox4minecraft/amidst/pull/700/files#diff-e81e43eadf5edbbf859559b0447dc95aL178): this may be more performant, but this really feels like a hack.

-----
> This PR is already close to done, and in my opinion this should also be merged in combination with @magicus's pull for caching previously rendered fragments

Magicus' PR hasn't been merged, because I saw unbounded RAM usage when moving the map around (see [this comment](https://github.com/toolbox4minecraft/amidst/pull/564#issuecomment-575737135)).

-----

The tests look encouraging, I got a ~3x performance improvement by using 4 threads (the number of physical cores on my machine), and a ~4x improvement by using all 8 virtual threads.

I've been able to consistently reproduce the broken fragments by doing the following by fully dezooming the map and "flinging" it into a corner, then stopping it suddenly. The few fragments which started loading while the map is moving doesn't seem to be invalidated, and display the biome (and structure) data for the wrong location.

> I have whittled the possibilities of the bug down to something with the BufferedImages.

The fact that the structure icons are incorrect leads me to think that the issue isn't specific to the `ImageLoader`s, but affect all the `FragmentLoader`s equally; I would suspect that the issue is somewhere in the `amidst.fragment.layer.*` classes.

burgerguy:
@moulins to address all of your comments:

BiomeProviders are probably a better option for long term support, but the the current implementation is, for lack of a better word, unfinished. I think the first thing we need to do is find a way to get rid of most of the reflection outside of class translation. For starters, unreflecting methods is really bad. It makes calling the method so slow it’s not even worth converting. It’s better to just use a Method rather than a MethodHandle. Second, we should find a better way to get the BiomeProviders. We can just get the constructor and make it directly without using the registry. Third, (I think) it’s cleaner just to get the Biome registry by getting the field from the Registry class. This method that I’ve been using requires no outside reflection. These three changes would make the current LocalMinecraftInterface a little more bearable to work with.

> I'm not fond of the change in BiomeDataOracle: this may be more performant, but this really feels like a hack.

Yea, sorry about that. It’s pretty disgusting, I have to agree. I thought I had already changed this back, but I guess not. When I get the chance, I’ll remove it.

> The fact that the structure icons are incorrect leads me to think that the issue isn't specific to the ImageLoaders, but affect all the FragmentLoaders equally; I would suspect that the issue is somewhere in the amidst.fragment.layer.* classes.

Hmm... It could be that but this lines up a lot with the two AtomicReferenceArrays in the Fragment class... I’ll look into it.

When I get home I’ll update more. Thanks for the feedback!

burgerguy:
I just realized I forgot to fix the temporary solution for detecting the BiomeProviderSettings

burgerguy:
I switched back to BiomeProviders, but with two key differences:

1. The unnecessary reflection was replaced with symbolic stuff.

2. We now get the biome registry directly through a field in the registry class. If this changes in the future, we can just change the real method name and it should work.

Along with this, I got rid of the weird method in the BiomeDataOracle and replaced it with detection of if the size is 1 in the getBiomeData() method, removed some unused stuff with the symbolic classes, and changed a bit of stuff in the processQueues() method. The only thing left I feel for this pull is to fix the main bug with fragment loading. After thinking about it for a bit, you’re probably right about it being something with the FragmentLoaders. I’ll mark this as ready for review after the bug is fixed.

PR #701, opened 2020-02-26
Add CPU usage widget

burgerguy:
Pretty simple. Shows the CPU usage of the program above the FPS widget.


moulins:
Good change, but I don't think this needs a separate setting. Why not put it inside the FPS widget?

Mockup:  
image
As a bonus, it would take a little less vertical space.

burgerguy:
I merged the CPU usage widget with the FPS widget as you said. Also, for later widgets, I added OffsetWidget which offsets the Widget from the specified anchor point.

moulins:
Looks good to me!

Issue #245, opened 2016-10-24
Add variable size photo export


SungMatt:
At the moment, photo exports are based on the size of your screen. This results in blurry exports of maps, and when printed, do not have as great quality. Addition of a variable high-resolution export would be helpful, for printing usages.

GeorgeWL:
Also,  could we have SVG image export? 

That way it scales near-infinitely up and down in size,  thanks to the images being a set of mathematical vectors.

stefandollase:
@GeorgeWL I will have to look into that. However, that should really be a separate feature request ;-)

moulins:
Closing as duplicate of #713.

PR #716, opened 2020-03-18
High quality biome image saving

### LINKED ISSUE #713: [Feature] Improved PNG Map Export Functionality
This may be a duplicate of #245, but I figured I could possibly make it more explicit.

The current export functionality takes a screenshot of current AMIDST screen area. This places an upper bound on the size of images that the exporter can produce, and means that zoomed-out exports have a lower resolution in terms of blocks-in-game.

Would it possible to alter the PNG export functionality to allow the user to specify:

- The rectangular coordinate area to map (either center-span or min-max for the _x_ and _z_ coordinates).

- (Optional) The resolution to export at, specified in the number of in-game-blocks per exported pixel.

- (Optional) Create a command-line argument to automatically take a screenshot (#711).

AFAIK, when AMIDST is first opened, the zoom level is such that one screenshotted pixel corresponds to four blocks in-game. Using autohotkey scripts, I've exported and stitched a large number of screenshots at this zoom level, and the results are quite spectacular compared to a native zoomed-out screenshot.

Finally, I would like to ask commenters **not** to turn this into a request for SVG export functionality (#327/#714). Although I am not opposed to this, the PNG format is probably better for storing the inherently rasterized data that AMIDST produces (with the exception of representing toggleable map-marker icons and grids). This is shown by the image below:

SVGdemo2

Note that although the raster-upscaled _circle_ is significantly degraded compared to the vector-upscaled one, the _rectangle_ suffers no degradation since it lies on the "pixel-grid" to begin with. The output of AMIDST (world biome data) similarly lies on a grid, and therefore can be represented using a (lossless) raster image format without _any_ loss of information/quality.

As an aside, converting an SVG image into a _**pixel-accurate**_ raster image isn't a trivial process (Inkscape, for example, antialiases on export by default.)

burgerguy:
Intended to eventually fix #713 ~~and #245~~

Currently:
* Allows exporting quarter resolution and full resolution biome data
* Exports what's on the screen
* Uses the TIFF format and a tiled approach to use less ram during creation and exporting
* Uses JAI to create and export the image
* Single threaded

Eventually:
* ~~Should be able to choose between a predefined size or just what's on the screen~~ _(Finished 3/19)_

* ~~Should be multi threaded~~ _(Finished 3/18 Note: Will do this after multithreading is implemented for biome data)_
* ~~Should export tile by tile for even less ram usage~~ _(Finished 3/18 Note: This works well enough with the tile cache in JAI)_
* ~~Shouldn't freeze the whole application~~ _(Finished 3/18)_
* ~~Should have a progress indicator~~ _(Finished 3/18)_

This will be a draft until further notice. Any comments or recommendations for features related to this are appreciated!

**Edit 3/18:** The implementation of the progress bar is kind of janky, but I think I flattened all of the possible bugs that would be caused due to it's inherent jankiness. It's kind of a band-aid solution, but it works.

**Edit 3/20:** I made the whole thing a lot cleaner, and implemented it as the WorldExporter. I didn't know if that was going to be used for something else, but I assumed it meant as and image and not as a world file. The only thing left to fix is a couple memory leaks that make it unable to export large images.


burgerguy:
I'm going to mark this PR as ready for review, as I have gotten the program to use much less memory than before and added all of the features that were initially planned. However, it still does use a lot of ram.

image
Using JVisualVM, I was able to see the ram usage during the steps of exporting the images. You can see a big spike at the exporting process, and I don't think I can cut this down any more unfortunately. Because of this, I changed the max ram used in the exe to be 2GB instead of 1GB so it doesn't run out of ram during this process.

[Here](https://www.easyzoom.com/imageaccess/061c1cde616d4adda22b0360c7d58c3e) is an example of what it looks like exported. It's full resolution, so you can zoom in and see all of the details. One pixel is equal to one block in game.

moulins:
The feature looks OK; the UI feels a little rough but this can be improved later.

However, I'm reluctant to bring a new third-party dependency for this. Did you test the memory usage when writing to a basic `BufferedImage` ? (Your example image is 12,353 x 9,290, so with 4 bytes per pixel I would expect "only" 500-600Mo of RAM to be necessary).

Additionally, it would be better to export to the PNG format if it doesn't increase the memory usage, to mimic the "Screenshot" functionality.

src/main/java/amidst/mojangapi/world/export/WorldExporter.java
```diff
@@ -1,49 +1,158 @@
...
+		private int threadNo;
+		
+		@Override
+		public Thread newThread(Runnable r) {
+			Thread t = new Thread(r);
+			t.setName("WorldExporterThread-" + threadNo++);
+			return t;
+		}
+	});
+	private static final WorkerExecutor exporterExecutor = new WorkerExecutor(pool);
```
moulins, line 46: I feel that this shouldn't create it's own executor, but instead use the one provided by `ThreadMaster.getWorkerExecutor()`, by passing it into the constructor.

Is there a specific reason why you allocated a specific thread pool for this?

src/main/java/amidst/gui/main/viewer/widget/ProgressWidget.java
```diff
@@ -0,0 +1,98 @@
...
+	public void setProgress(int progress) {
+		this.progress = progress;
+	}
+
+	@Override
+	protected boolean onVisibilityCheck() {
+		return (progress >= max || progress < min) ? false : true;
+	}
+	
+	public void acceptProgress(String progress) {
```
moulins, line 78: This should use the `ProgressEntryType` enum defined below, together with the `int` value, instead of using a `String`.

src/main/java/amidst/mojangapi/world/export/WorldExporter.java
burgerguy: I have this creating it's own executor so we could have a reliable implementation for the static method `isExporterRunning()`. I tried experimenting with seeing if a thread exists by name so we wouldn't have to do this, but I couldn't reliably get if it was running at the time or not. We may be able to extend Thread and have a volatile flag that says if the thread is active or not that is run before and after the thread executes, but that seems kind of janky.

src/main/java/amidst/gui/main/viewer/widget/ProgressWidget.java
burgerguy: I had to wrap the two in an `Entry<>` for it to be compatible with the current progress reporting code, but hopefully it's a bit better than before.

src/main/java/amidst/mojangapi/world/export/WorldExporter.java
moulins: Why not use an static `AtomicLong` field to replace the `pool.getActiveCount()` method?  
Something like:
```java
private static AtomicLong ACTIVE_THREADS = new AtomicLong(0);

public void export() throws WorldExportException {
	exporterExecutor.<String> run(() -> {
		ACTIVE_THREADS.incrementAndGet();
		try {
			this.doExport();
		} finally {
			ACTIVE_THREADS.decrementAndGet();
		}
	}, progressListener, this::onFinished);
}
```

burgerguy: Just implemented your solution. I don't know why I didn't think of that earlier, but it's a lot simpler.

burgerguy:
@moulins The reason we use the TIFF file format is so we can save the file in chunks. PNG doesn't support tiled images like TIFF does, so we use TIFF instead. Also, when I tested it, exporting a regular `BufferedImage` with a size of 20000x20000 used up over 8GB of ram during exporting. I think this is due to it saving the raster in multiple places at once, or some of it not being wiped by the garbage collector efficiently. We basically have to use tiled images if we don't want to load the whole thing into ram, which quickly becomes a problem with full resolution exports. Also yea, the UI has some flaws. At some point I'd like to merge all 3 popups into a more organised window so it's not as annoying. On the other hand, I think the progress bar is fine and the position of it in the window isn't too bad.

burgerguy:
Just finished working on a much nicer UI. When you export a biome image, it will prompt you with this:

Export-GUI

Most of the settings are self explanatory. The coordinates for top, bottom, left, and right are what they would be in the amidst viewer window (lower values correspond with up, higher with down, etc). The preview is 100 pixels x 100 pixels and is generated on another thread. It is scaled up to 200x200 in the GUI.

burgerguy:
I'm changing this to a draft because I need to refactor some bad code and tweak some things.

moulins:
The new UI looks good! Is the preview regenerated when the user changes the bounding box?

---

In other news, I may have found a solution to the memory usage: use a custom [RenderedImage](https://docs.oracle.com/javase/7/docs/api/java/awt/image/RenderedImage.html) which computes the pixel data as the encoder requests it, so that we never need to manipulate the whole image.

I'll have to write a proof-of-concept to see if it works like I think it should, but it looks promising.

burgerguy:
I'll have to check out that RenderedImage to see if it can fit. We still will probably have to stick to TIFFs so it doesn't have to load it all into ram. With a custom implementation of RenderedImage, we can export and generate at the same time. I don't know how the ProgressBar will fit into this yet, but I'm gonna have to make some changes to how that works as well. In the meantime, I've cleaned up a ton of code and fixed some bugs. The preview does change when the bounding box is changed or if the full resolution checkbox is clicked. The bounding box defaults to the main window size and position when you open the window. I made the full resolution checkbox work on the preview, but you can really only see the difference with a small bounding box. It doesn't impact performance much, so I'm probably going to keep it that way.

moulins:
> I looked into RenderedImage and tried to create a dummy implementation that references its data to the BiomeDataOracle, but it required creating a raster during export. This defeats the purpose of having something like this, so I think that the current implementation works best unless you can come up with an implementation that works better.

I've made my own (very simple) test [here](https://gist.github.com/moulins/b5256ab5640b738de7ff1ac58ca4b042). Even when recreating a new `Raster` each time, I still get very good memory usage (with a 10000x10000 image):

image

My code fill each returned `Raster` with a random color, and returns images like this:
image
Which means that the PNG encoder requests the data line-by-line.

This doesn't seem to be an ideal access pattern for computing biome data, so we may still want to compute biome data in larger strips (maybe 16 or 32 pixels tall?).

burgerguy:
I'm starting to wonder how horribly ram inefficient `BufferedImage` is. At least we found something that works, but I feel like something's off here that we aren't accounting for that will show up when implemented. Also, when we don't use a disk cache, we end up hitting a theoretical maximum for how big of an image we can export. We may want to put a warning if it thinks that an image will max out the ram.

burgerguy:
I was able to get the `RenderedImage `to work in the program, and it works extremely well. It uses barely any ram compared to the old `TiledImage` approach, and doesn't require an external library. I think I caught all the bugs that needed fixing, and while the `CustomRenderedImage `is a little janky, everything else has gone through atleast 2 iterations of cleaning. This pull request is probably finished unless I see any more bugs, because I think everything that was needed is done.

moulins:
I've done some testing on my side and didn't see any bugs either, so everything seems good!

I'll merge this as soon as you've resolved the conflicts with the latest `master` changes ;)

moulins:
Thanks for the work!

Issue #242, opened 2016-10-14
Superflat and Amidst


XLegacYX:
I notice that AMIDST doesn't have a map where it is completely flat with the one biome you have in a world. Instead, it just shows lots of biome's like a normal World map would look like (but, some structures actually were correct, but still).

So, I recommend that you could add a feature to have a way for AMIDST to show superflat land with the one biome.

stefandollase:
Thanks for the report. This seems to be a bug. Probably the *flat* world type in Amidst is supposed to display a Superflat world. Right now, I cannot tell what is the cause of this bug. We will have to investigate this further.

XLegacYX:
I didn't realize it was bug until you told me. :P
Well, hope you can fix it.

PR #720, opened 2020-03-25
Add support for Minecraft 20w11a and higher

moulins:
Add support for the latest Minecraft snapshots (20w11a and 20w12a).

As usual, I'll wait a bit for new snapshots before merging this, to make sure that it is reasonably robust.


breon:
Any updates on this? I pulled this branch and built from source (mac), but I get a crash. Mojang is up to 20w17a.

moulins:
Yes, this branch broke around 20w15a, I think. :frowning_face: (this is why I didn't merge it immediately)

I'll try to look into it next week.

moulins:
The last two commits add support until the current snapshot, 20w02b.

Mojang refactored away the class we were using to create the world and instanciate the biome providers, so I was forced to find another way and create a new `LocalMinecraftInterface`. Let's hope this doesn't break in the next snapshot! :D

---------------

There was an unexpected benefit however: the new way of creating the world is much cleaner, and actually fixes a long-standing issue regarding Superflat worlds (#242)!

burgerguy:
I'm pretty sure my refactoring from a while ago already used the cleaner world creation you're referencing, if the class you're talking about that was removed was the one that got the BiomeProvider instances from the registry. That was actually really annoying to work with as complying with Minecraft's registry through reflection is such a pain you might as well instantiate the objects you want yourself. Unfortunately with things like Biomes, we are forced to go through the registry, but I think for everything else we should just directly create them.

moulins:
I'm pretty sure this is a different method, given that I'm using a class introduced in `20w20a` (`WorldGenSettings`). This class provides a `read` static method taking a `java.utils.Properties` instance, and which returns an object containing generators for all the dimensions.  
This means we don't need to fiddle with `BiomeProviderType` or `BiomeProviderSettings`, all is done by the Minecraft code.


`MinecraftInterface` code [here](https://github.com/toolbox4minecraft/amidst/blob/18d829a61110ebfa7e054956c73acdba99a4b6d0/src/main/java/amidst/mojangapi/minecraftinterface/local/LocalMinecraftInterface.java#L92).

PR #724, opened 2020-03-27
Fix launcher profile bug

burgerguy:
Fixes a bug that didn't allow you to open a launcher profile twice.


moulins:
Good catch, I forgot to close the zip FileSystem after working with it!  

However, your fix is still wrong, because it would leak file handles; I implemented a proper fix in ae81cf5 instead.

burgerguy:
When I tried to close the `ZipFileSystem` manually with a try finally and `.close()`, it threw a `ClosedFileSystemException` when I tried to make a new one with the same path later. I assumed that it wasn't supposed to be closed, but try with resources seems to do it without any problems. Weird.

moulins:
Yeah, I could use the same profile several times without relaunching Amidst, so it works!

PR #730, opened 2020-04-07
Small changes that may help out in the future

burgerguy:
* Fix a method in RealClass that sometimes didn't work as intended
* Removed the class maximum file size as it wasn't detecting some classes in b1.7.3, replaced it with seeing if the file is a class file
* Add an option for symbolic and real arrays in class detection
* Fixed the magic string for b1.9-pre1


moulins:
Note: I've reverted your change to `RealClasses`, because 1.15.2 wasn't loading anymore.

burgerguy:
It works fine on my end but we probably don't need it until we either add beta 1.7.3 support or some update comes out and messes it up.

PR #742, opened 2020-04-23
Update WorldSeed.random()

burgerguy:
Beforehand I was under the impression that SecureRandom.nextLong() only could generate 2^48 results because it directly inherited it from Random. I have updated it now, and added comment in the method:

> SecureRandom.nextLong() can generate 2^64 different results even though it directly inherits it from Random. The key difference is the nature of next() in the different classes. In Random it is restricted to 2^48 bits, while SecureRandom uses (i think) 2^128 bits.


moulins:
Thanks! I've cherry-picked your commit into `master` (5a37f67b) to avoid the empty merge commit.

PR #748, opened 2020-05-02
Added catch for HeadlessException and comprehensive error message

mawenzy:
I'm using Fedora 31 and Java comes preinstalled with openjdk-headless, which allows me to run amidst, but it crashes when trying to create a `JFrame` with an HeadlessException.

I added an error message offering a probable solution, installing the not headless version.


moulins:
The message is good, but I don't think the check belongs here.

A better way would be to check the [GraphicsEnvironment.isHeadless()](https://docs.oracle.com/javase/7/docs/api/java/awt/GraphicsEnvironment.html#isHeadless()) flag, just before the application starts (for example, in the `Amidst.run()` method).

I'll accept the PR if you make these changes :)

mawenzy:
Makes sense, I have changed it.

Issue #369, opened 2018-02-04
Extensions for starting with custom MinecraftInterface


fraenkelc:
I've hacked together a working integration of Amidst into a Forge client ( see https://github.com/fraenkelc/amidst-forge ). The interfaces made it straight forward to call the real BiomeProvider / GenLayer code and use that. What gave me some issues was supplying that custom `MinecraftInterface` instance to the amidst client. What I've come up with can be seen in https://github.com/fraenkelc/amidst/commit/a5dca69336771ef4805ff745af4d908fbc062ec1 .
It's neither pretty, nor complete; I'd like to have the option of disabling the profile switcher and load from save game menu items for example.

Is there any interest in creating / integrating an interface for that usecase? I thought I'd raise the question here before doing more work on the amidst codebase.

stefandollase:
Thanks for your effort, I like the idea. However, I am a bit afraid to put Amidst into a new execution context, since I am not familiar with forge and I am sure there will be bugs which are related to putting Amidst into the new execution context. Any thoughts about this?

fraenkelc:
Embedding amidst and running it in the same process just was the easiest way to get the integration running and it required the least modifications to amidst. You're right, there's various things forge does that might cause issues, such as paching minecraft and java classes, using the minecraft logger and also registering its own uncaught exception handler.

A different way of integrating amidst with a modded client could be achieved using a remote connection with a lightweight binary protocol. If that's too slow memory mapped files might be an option. That way Amidst could run in its own JVM. This approach would require some extensions to the MinecraftInterface to handle updates to the biome list.

fraenkelc:
Time to give this a poke a year later. There is some interest in running amidst against modded instances - the most recent version of my fork had 1800 downloads in the last 4 months.

 Is there any interest in adding a remoting interface and extending amidst to accept a biome list and biome data from a running modded minecraft instance? If so, how can we move forward on that?

moulins:
Having the capability to work with Forge profiles would be a welcome addition to Amidst, but − as you've seen yourself − there's lots of work to be done in the Amidst codebase for this to be possible.

A remote interface seems the way to go for the long-term, but before implementing it, I think the `MinecraftInterface` should be improved, and become the source of truth for the following:
 - the `Biome` list (so that worlds with custom biomes are correctly displayed);
 - the list of compatible biomes for each structure supported by Amidst (so that structure locations are correct on worlds with custom biomes);
[...]

PR #750, opened 2020-05-04
Add support for multiple worlds on a single MinecraftInterface

moulins:
This refactors all `MinecraftInterface`s to allow creating multiple worlds at the same time with a single `MinecraftInterface` instance. As a consequence, the Amidst world disposing logic (via `World.dispose()`) isn't necessary anymore, and has been removed.

Additionally, it should now be possible to switch worlds while an export operation (see #716) is in process; I still need to do more testing to be sure that nothing breaks before making this change.

### DIFF
--- a/src/main/java/amidst/gui/export/BiomeExporterDialog.java
+++ b/src/main/java/amidst/gui/export/BiomeExporterDialog.java
@@ -1,5 +1,14 @@
 package amidst.gui.export;
 
+import static java.awt.GridBagConstraints.BOTH;
+import static java.awt.GridBagConstraints.CENTER;
+import static java.awt.GridBagConstraints.EAST;
+import static java.awt.GridBagConstraints.HORIZONTAL;
+import static java.awt.GridBagConstraints.NONE;
+import static java.awt.GridBagConstraints.SOUTH;
+import static java.awt.GridBagConstraints.SOUTHEAST;
+import static java.awt.GridBagConstraints.SOUTHWEST;
+
 import java.awt.AlphaComposite;
 import java.awt.Color;
 import java.awt.Component;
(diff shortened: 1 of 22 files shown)


burgerguy:
Off of this branch I have created a fully working implementation of the multithreaded fragment loading. Once this eventually gets pulled I'll start a new PR, but for now I'm going to take down the old one, #700.

moulins:
On #369, @fraenkelc proposed the following improvement:
> For [a future] remote interface the World interface could use a close() method so that I can also discard the world and resources in the remote minecraft instance once they are no longer used in amidst as garbage collection will not work in this case (across processes).

I'm not sure this can be easily added, because explicit lifetime management in a multi-threaded context is hard, but I'll put this PR temporarily on hold to see if a can find a good solution.

burgerguy:
@moulins We could call the `close()` method on `finalize()` so it closes when the garbage collection happens, but I haven't tested this.

On a similar note, there are a lot of things that should be getting GC'd but aren't in the current state. This includes `RunningLauncherProfile`s, `MinecraftInterface`s, `FragmentQueueProcessor`s, and possibly `FragmentManager`s. There are probably more that I missed, but I think it's partially caused by `ViewerFacade.dispose()` and `Application.dispose()` not doing nearly as much as they should. Some things are staying in memory after `Application.restart()`, and it becomes pretty obvious that there's a memory leak when the garbage collector starts to pause for longer and longer periods. I'll look into fixing all of this, but it's gonna require a lot of refactoring to be done right, and the addition of allowing multiple worlds per MinecraftInterface isn't helping in this aspect when they aren't properly disposed anymore.

moulins:
> We could call the close() method on finalize() so it closes when the garbage collection happens, but I haven't tested this.

This *could* work, but doing actual work in a `finalize()` method is generally a bad idea, because:
  - There is no guarantee that the method will ever get called by the garbage collector;
  - The method is called synchronously by the garbage collection thread, so something like a network call (which could potentially take several seconds) could completely stall the application;
  - The `finalize` method is deprecated :D (there is a replacement with `PhantomReference`s I think, but it's difficult to use).

I think the actual solution will be to mark `MinecraftInterface.World` as *not* thread-safe (and refactor so that each thread creates its own World, for exemple for the biome export), so that the lifetimes can be cleanly managed.

PR #754, opened 2020-05-08
Add support for language files

burgerguy:
While most of the code involved with using the language files is done, I'm still going through the process of actually moving most of the current strings to the en-US language file. When it's finished though, we should be able to start working on translations for other languages.


moulins:
This is a good initiative, but I would prefer waiting for Minecraft 1.16 and the next release of Amidst before starting these kinds of cross-cutting changes.

There is already a lot of changes since v4.4, and I don't want to cram too much features into a single release.

--------

I'll close this PR for now, but don't hesitate to create another one as soon as Amidst v4.5 is out!

PR #756, opened 2020-05-13
Profile select enhancements

burgerguy:
* Adds support for profile icons, including custom ones.
* Makes the scrollbar faster so it's not as slow to scroll through the pane.
* Added a color for when a failed profile is selected.
* Profiles are now sorted alphabetically.

The files included for the profile icons are the ones found in the [launcher.dll](https://launcher.mojang.com/v1/objects/d4f3d5bbd0cf31ba8006452bd54e45715ebff23f/launcher.dll) file. I couldn't find a way to just download them from there because the DLL is compressed with LZMA. It also probably isn't a good idea to download it as it takes up 16 MB, which is many times the size of amidst. I couldn't find the icons anywhere else in the assets, so I don't think it's possible to download them separately. We may be able to just use a part of the block texture associated with it, but that's kind of a last resort.


src/main/java/amidst/gui/profileselect/ProfileComponent.java
```diff
@@ -61,24 +62,33 @@ private void updateIfNecessary() {
...
+					if (widthSum + xOffset > maxWidth) {
 						return result.substring(0, i) + "...";
 					}
 				}
 			}
 			return result;
 		}
+		
+		@CalledOnlyBy(AmidstThread.EDT)
+		private void drawProfileIcon(Graphics2D g2d) {
```
moulins, line 85: The `scaledIcon` should probably be cached, to avoid reloading it each frame.

moulins:
Good improvement! Having the profile icons will certainly help those who have lots of profiles.

> I couldn't find the icons anywhere else in the assets, so I don't think it's possible to download them separately. We may be able to just use a part of the block texture associated with it, but that's kind of a last resort.

Yeah, this seems very complicated for a marginal gain, as the profile icons only take a few tens of kilobytes. Simply including them in the JAR file is probably the best solution.

burgerguy:
@moulins I was more worried about the possibility of being accused of stealing Mojang's assets. Sort of like how we don't use any Minecraft code and instead work around it, I was thinking about doing the same here. Do you think that is something we should be worried about, or is it ok?

src/main/java/amidst/gui/profileselect/ProfileComponent.java
burgerguy: Just fixed this I think.

moulins:
> I was more worried about the possibility of being accused of stealing Mojang's assets.

Given that the Minecraft wiki reuses a lot of Minecraft assets and never had issues, I would be surprised if this happened (and in the worst-case scenario, we can just remove the feature).

PR #759, opened 2020-05-16
Add End Gateways and speed up some things in EndIslandOracle

### LINKED ISSUE #116: Map the End gateway portals, if possible
[from reddit](https://www.reddit.com/r/Minecraft/comments/48e3he/amidst_for_19_now_available_locate_igloos/d0m6m86)

> Suggestion: Could you give the locations where the outer Gateways generate, perhaps with a line linking them to the corresponding spot around the center island and an indication of which order they appear in?

> My understanding is that you can create up to 20 gateways, and the inner ones make a circle, but where the outer ones appear is not as easy to know in advance. Can AMIDST figure it out?

I don't know whether it can be determined by the seed, but seems worth looking into.

### LINKED ISSUE #234: 1.11 "Escape portals"
New in 1.11, escape portals that lead to the End spawn platform will generate (randomly?) near End Cities.

These are similar to the End Gateways discussed in #116 and would likely use the same layer/icon

burgerguy:
Fixes #234 and fixes #116

This doesn't add the outer end gateways that are generated when you throw an ender pearl into a gateway on the main island. Those were going to be hard to do and it was pretty unrealistic. This does include the ones on the islands and the ones randomly generated around the end.


burgerguy:
I got the outer ring gateways to show an approximate location of where they will generate. Usually they end up ~40 blocks away from where they actually are, but we can't really do anything about that. It does take a few seconds to generate the cache, but afterwards it doesn't cause any slowdowns.

I also fixed the cursor information for biomes in the end. They now show the correct biomes instead of just one.

grit96:
Looks like the tests are failing at `Running amidst.mojangapi.world.test.TestWorldEntryEquality`.

grit96:
@burgerguy I tested this on v4.5-beta1 and it works perfectly.

burgerguy:
@grit96 I'm still working on some other changes to the end at the moment, so I'll mark this PR as a draft. Also with the tests, I'm not sure what's causing them even after looking through them.

burgerguy:
I marked this PR as ready for review after fixing up a few things. First, the stuff that was already there was updated to work with 1.16. Second, I added the small end islands starting from 1.13. While it is true that it wasn't possible to procedurally generate the small end islands from the seed alone, 1.13 allowed that to be possible.

Screenshot:
image

burgerguy:
@moulins This PR should be ready for merging. It's up to date with master and includes everything that's needed.

burgerguy:
I figured out why the tests weren't working. Due to the refactoring of the end islands, I had to make some changes to the `TestWorldEntryEquality` test. It still tests the large islands only, so maybe at some point that could be changed to work with the small islands too.

PR #772, opened 2020-06-19
Add multithreaded fragment loading

### LINKED ISSUE #521: Render using multiple threads.
Currently it looks like amidst renders one chunk (?) at a time in a single thread.  On a multi-core CPU rendering could be done more faster if multiple threads were used.

burgerguy:
I think I've finally been able to flatten out most of the bugs with this implementation. I was able to fix a lot of memory leaks in the process, and I also made using 1.14 stop the useless datafixer threads.

EDIT: this fixes #521


burgerguy:
@moulins I just fixed the branch and updated it for 1.16. All of the conflicts are resolved and all of the possible memory leaks that I could find have been patched out, and I think it's ready for review.

src/main/java/amidst/fragment/Fragment.java
```diff
@@ -18,26 +19,29 @@
...
- *
+ * life-cycle consists of the three flags: isInitialized, isLoading, and
+ * isLoaded. isInitialized can be set to true from any thread, however setting
+ * isInitialized to false as well as any modification to isLoading and isLoaded
+ * will always be called from the fragment loading thread, to ensure a
+ * consistent state. Also, isInitialized will only be set to true again after it
+ * was set to false. It is not possible that isLoading or isLoaded is true while
+ * isInitialized is false. It is also not possible for isLoaded to be true while
+ * isLoading is true. </br>
+ * </br>
```
moulins, line 35: This feels quite confusing, and I think these three flags actually want to be a state machine.

From your description, there are only four valid states:
 - `!isInitialized && !isLoading && !isLoaded`; the fragment is uninitialized
 - `isInitialized && !isLoading && !isLoaded`; the fragment is waiting to be loaded
 - `isInitialized && isLoading && !isLoaded`; the fragment is loading
 - `isInitialized && isLoading && isLoaded`; the fragment is fully loaded

It is possible to regroup the three flags into a single `state` field, which can take one of the four values given above?

src/main/java/amidst/mojangapi/minecraftinterface/MinecraftInterface.java
```diff
@@ -59,7 +62,23 @@
...
+	 * WorldAccessor can.
+	 * 
+	 * This also includes the createWorldAccessor method so it's required to create
+	 * a WorldConfig first.
+	 */
+	public static interface WorldConfig {
+		
 		public Set<Dimension> supportedDimensions();
+		
+		public WorldAccessor createWorldAccessor(long seed, WorldType worldType, String generatorOptions) throws MinecraftInterfaceException;
```
moulins, line 82: I'm not sure about this `WorldConfig`/`WorldAccessor` split; I don't really see the point of the `WorldConfig` objects, given that in all `MinecraftInterface`s implementations these objects are basically empty.

The only benefit is that the initialisation code can be run separately, but why can't this be bundled with the creation of the first world?

src/main/java/amidst/mojangapi/minecraftinterface/legacy/LegacyMinecraftInterface.java
```diff
@@ -63,33 +62,26 @@ public LegacyMinecraftInterface(Map<String, SymbolicClass> symbolicClassMap, Rec
...
-		if (isIntCacheInUse) {
-			throw new IllegalStateException("This thread is already generating biome data");
-		}
-		isIntCacheInUse = true;
 		try {
 			resetIntCacheMethod.invokeExact();
-			return (int[]) getIntsMethod.invokeExact(biomeGenerator, x, y, width, height);
+			int[] biomeInts = (int[]) getIntsMethod.invokeExact(biomeGenerator, x, y, width, height);
+			// we have to clone the array so we aren't using a reference to an array that's still in the IntCache
+			return biomeInts.clone();
```
moulins, line 82: Instead of cloning the `biomeInts` array, it would be more efficient to take the `biomeDataMapper` as a parameter and call it's `apply` method directly. This should be thread-safe, given that we are in a `synchronized` block and nobody else can access the `biomeInts` array.

src/main/java/amidst/util/SwingUtils.java
```diff
@@ -0,0 +1,91 @@
...
+	 * This helps with allowing swing components to get garbage collected.
+	 */
+	@CalledOnlyBy(AmidstThread.EDT)
+	public static void destroyComponentTree(Component c) {
+		// calls the component's dispose method if it has it.
+		try {
+			Method m = c.getClass().getMethod("dispose");
+			m.invoke(c);
+		} catch (Exception e) {
+		}
```
moulins, line 26: I think only `java.awt.Window` has a `dispose()` method, so this could probably be replaced by an `instanceof Window` check.

src/main/java/amidst/util/SwingUtils.java
```diff
@@ -0,0 +1,91 @@
...
+						// that doesn't extend from EventListener. If so, ignore it
+						System.out.println("Listener " + params[0] + " does not extend EventListener");
+						continue;
+					}
+					for (int j = 0; j < listeners.length; j++) {
+						try {
+							method.invoke(comp, new Object[] { listeners[j] });
+							//System.out.println("removed Listener " + name + " for comp " + comp + "\n");
+						} catch (Exception e) {
+							System.out.println("Cannot invoke removeListener method " + e);
```
moulins, line 72: This should be `AmidstLogger.info(...)`

src/main/java/amidst/util/SwingUtils.java
```diff
@@ -0,0 +1,91 @@
...
+							// continue freeing listeners to make sure we can free as much
+							// memory has possible
+						}
+					}
+				} else {
+					// The only Listener method that I know of that has more than
+					// one argument is removePropertyChangeListener. If it is
+					// something other than that, flag it and move on.
+					if (!name.equals("removePropertyChangeListener"))
+						System.out.println(" Wrong number of Args " + name);
```
moulins, line 85: This should also be `AmidstLogger.info(...)`

src/main/java/amidst/util/SwingUtils.java
```diff
@@ -0,0 +1,91 @@
...
+				}
+			}
+		}
+	}
+
```
moulins, line 90: This code uses 8 spaces indentation; this should be fixed to match the rest of the Amidst code.

src/main/java/amidst/fragment/Fragment.java
burgerguy: This has been fixed by creating a new enum in the `Fragment` class called `State`. There are now four different states: `UNINITIALIZED`, `INITIALIZED`, `LOADING`, and `LOADED`.

src/main/java/amidst/mojangapi/minecraftinterface/MinecraftInterface.java
burgerguy: I removed all of the `WorldConfig` stuff and instead changed some stuff in `ThreadedWorldAccessor` as well as `LoggingMinecraftInterface` so it works like it used to while still having multithreading.

src/main/java/amidst/mojangapi/minecraftinterface/legacy/LegacyMinecraftInterface.java
burgerguy: Fixed this by doing what you said.

src/main/java/amidst/util/SwingUtils.java
burgerguy: Fixed. I could see the possibility of a custom component having a method called `dispose()`, but I guess it would be the job of that component to do it itself anyway.

src/main/java/amidst/util/SwingUtils.java
burgerguy: Fixed

src/main/java/amidst/util/SwingUtils.java
burgerguy: Fixed

src/main/java/amidst/util/SwingUtils.java
burgerguy: Fixed.

moulins (approved):

PR #803, opened 2020-07-03
Implement better interpolation algorithm + other minor changes

burgerguy:
Previously, the interpolation algorithm was set to interpolate bilinearly when the scale was less than 1, and nearest neighbor otherwise. This changes it so it instead does a bilinear nearest-neighbor hybrid. Whenever it detects that it can use nearest neighbor without changing pixel locations, it sets it to that. Otherwise, it will split the scale value and render it partially in nearest neighbor, and partially in bilinear. This gets rid of the flickering that you'd typically see when zooming in with nearest neighbor, while still keeping the quality that nearest neighbor delivers.

This also changes the interpolation algorithm for the icon widget and the world icons. Now, when a world icon is selected, it switches to bicubic. Otherwise, it's nearest neighbor. For the icon widget, it's been set from nearest neighbor to bilinear because it was scaling them strangely before.

This doesn't seem to impact performance much, and while the approach isn't thread safe due to the reuse of a single image as a buffer, it shouldn't matter because drawing should only be called by the EDT anyway. I also made sure to use a `VolatileImage` rather than a `BufferedImage` because, while `BufferedImage`s can be accelerated, they're much harder to get to accelerate than `VolatileImage`s. When I started with trying to use a `BufferedImage`, it hurt performance by quite a bit due to the lack of acceleration.

I also added Pixel Scale to the debug widget. This allows you to see precisely what zoom you're at to see if it should be set to nearest neighbor or not.


moulins:
Looks good, merging.

PR #823, opened 2020-07-20
Add search bar to biome widget

### LINKED ISSUE #822: Feature request: Biome list filter
The biome highlight list keeps growing bigger and bigger. It would be very nice to have a search bar that filters that list. So, for example, if I write "ocean" in a search bar, I should only see the ocean biomes in that list. Everytime I want to look at something, it takes a lot of time to go through the list and find the biomes I'm looking for to highlight.

burgerguy:
Fixes #822.

Preview:
image


Boobies:
Any reason why this hasn't been pulled yet? I'd love to see it in the next release.

burgerguy:
This PR hasn't been merged because it does some janky things to make the text field work. I think @moulins wants to do a whole Widget refactoring before adding this, but I could be wrong.

PR #827, opened 2020-07-25
Fix full resolution biome exporting

### LINKED ISSUE #808: Biome map resolution lower than actual world map even when full resolution is selected
Hello,

The blocks on the biome map seem to be much larger than the actual biome granularity. I exported the biome map from VoxelMap and that has block level granularity where as amidst seems to have 4x4? block level granularity. Maybe this is really a feature request but I didn't see any mention of the biome map being low resolution. 
What I expected to happen was seeing the full granularity when I selected full resolution, what I actually saw was 

I have created an overlay of the VoxelMap export and the corresponding amidst biome map export. 
With the overlay
biome_map_overlay
Without overlay
biome_map_no_overlay
I am running on windows, with minecraft 1.16.1 using amidst v4.5 beta3, java version "9.0.4", though this seems to be present regardless of minecraft/amidst version. 
the seed I am using is -676194426 and the area in question is just north of 0,0 
The error can be reproduced simply by exporting the map at full resolution. the VoxelMap can be produced by using that mods map export feature while in biome display mode.

burgerguy:
Fixes #808, and may possibly fix a few incorrect structure positions for versions between 1.13 and 1.15.2.

### DIFF
--- a/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java
+++ b/src/main/java/amidst/mojangapi/minecraftinterface/local/DefaultClassTranslator.java
@@ -64,7 +64,7 @@ public enum DefaultClassTranslator {
                 .ifDetect(c -> !c.getRealClassName().contains("$")
                     && c.getRealSuperClassName().equals("java/lang/Enum")
                     && c.hasMethodWithRealArgsReturning("long", "int", "int", "int", null, null)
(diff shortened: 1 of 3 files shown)


moulins:
Thanks for the fix! I've converted the added version checks into proper `VersionFeatures` (09297267), and also determined the actual Minecraft versions which introduced the changes (1739bcdd)

PR #830, opened 2020-07-26
Retain biome selection between world switches

### LINKED ISSUE #824: Highlight Biomes
it keep resetting, 
in the previous version , the highlight biomes i choose is stay when i click new seed

i hope u get the point
*english not good

burgerguy:
Fixes #824. The biome selection widget may have been intended to reset when the world switches, but I'm not sure.


moulins:
I've just pushed 8b4e07ba to fix #824, so I'll close this.

PR #839, opened 2020-08-06
Fix integer overflow in end city icon generation

### LINKED ISSUE #838: End city icons no longer show after roughly 740,000 blocks from center
As displayed in the image below, end city icons stop showing after about 740,000 blocks from the center island. Seems to happen in a circle around the center island, so this number is possibly a radius. Can easily be reproduced with any random seed on either axis using `World` -> `Go to Coordinate`. Tested on Java 1.8.0_261 under Windows 10 Pro (Version 1909).

image

Deltanic:
Fixes #838.

Would likely make more sense to pass chunks along as a `long`, which was also mentioned in `StructureProducer`: https://github.com/toolbox4minecraft/amidst/blob/57444ba1268523102687e1e7b6b2057e39c1e3b5/src/main/java/amidst/mojangapi/world/icon/producer/StructureProducer.java#L53-L75

Perhaps that's a change worth making?


moulins:
Thanks for the PR!

PR #845, opened 2020-08-12
Add support for 1.16.2

burgerguy:
Also made all of the `MethodHandle`s use `invokeExact` by erasing their types for performance.


moulins:
I've been working on 1.16.2 and just pushed f644cd9, and didn't see that you had made a PR to fix the issue :').

I'll close this, and also add your `MethodHandle` modification in another commit if I can reproduce the performance increase.

burgerguy:
The performance increase is more noticeable in later Java versions. While it may not show much of a difference in Java 8, it may show a difference in a newer version like Java 12. Also, it only adds a single line to the code. What reason is there to not to keep it in if we went through the hassle to switch from `Method`s to `MethodHandle`s in the first place?

burgerguy:
@moulins Is there a better way to contact you so that we can avoid situations like these? It would save a lot of time and it would be nice to gain more insight into the project and what direction you're wanting to bring it towards. I'm always available on discord (burger#0007) but if you're unable to or would rather use something else let me know.

moulins:
Good point regarding the Java version, I only test on Java 8; I'll cherry-pick the commit then.

> Is there a better way to contact you so that we can avoid situations like these?

The best way would be to make a comment on the relevant issue (or create a new one) and @mention me when you intend to do substantial changes. I've also sent you a friend request on Discord so we can talk here if you want.

PR #885, opened 2020-08-25
Use monitor's refresh rate

burgerguy:
This sets the refresh rate of the render and queue processing threads to the same as the refresh rate of the best monitor connected.


burgerguy:
I marked this as a draft because some things like scrolling speed depend on refresh rate rather than time elapsed, so I'm going to have to fix those before this gets unmarked again.

moulins:
This is a good idea :)  
Maybe the framerate should be capped at a "reasonable" value (like 60Hz or 90Hz), to reduce CPU usage when the monitor has a high framerate, like 144Hz ?

PR #905, opened 2020-10-01
Added list of future feature support requested by the community

ParthMakode12345:


moulins:
The list of user-requested features is already available by filtering issues with the `feature` tag.

Adding such a list to the README is redundant information, and would additionally need to be kept up-to-date.

PR #935, opened 2021-01-16
Support all betas

kahomayo:
This PR adds support for versions a1.2.0 to b1.7.3 to Amidst. This includes displaying oceans at a reasonable accuracy, concluding the work that was started four years ago on the `support_early_betas` branch. Additionally, this PR also fixes a bug in Fragment rendering that would sometimes cause rendering to stop after switching worlds.

I encourage you to read the commit messages for details, so I will just be providing a brief overview in this PR message.

Most biomes can be determined by just using Minecraft's BiomeGenerator, which is fast. Oceans are not included in that output, however, and a view that doesn't include them is often inadequate. Oceans are determined when Minecraft shapes the chunk, so this code runs that method. Just doing this would be glacially slow, so I added an optimized perlin octave noise implementation that gets injected and provides very reasonable performance.

One optimization was to limit the number of octaves being used in the 3D interpolation noise. I didn't feel comfortable adding a configuration option just for beta support (and also didn't feel like doing GUI stuff), so the `OCEAN_PRECISION` constant is currently hard-coded to `4` (out of 16). Below is a full-resolution 5000x5000 comparison of the seed 'kaai'. [This is how it looks at precision 16](https://user-images.githubusercontent.com/74737340/104809655-0b1a3800-57ef-11eb-8d72-528559379d1f.png).

The comparison images were edited such errors in low-precision output are highlighted. Missing land is lilac, missing ocean is yellow. All images are compared against the full-precision output linked above. Note that the biomes obtained from Minecraft's BiomeGenerator were always identical.

[0 Precision](https://user-images.githubusercontent.com/74737340/104809657-0bb2ce80-57ef-11eb-84c7-f3c3ba8dcb4d.png)
[2 Precision](https://user-images.githubusercontent.com/74737340/104809651-09507480-57ef-11eb-8dd3-3901d61bdaae.png)
[4 Precision](https://user-images.githubusercontent.com/74737340/104809653-0a81a180-57ef-11eb-9e09-73ae0115b708.png)

~~This PR adds a dependency on Byte Buddy which is required for displaying oceans. The exact uses are outlined in the commit "Implement oceans in beta terrain".~~

The Byte Buddy dependency has been removed in the commits "Match overworld dimension" and "Include ocean generation".


src/main/java/amidst/mojangapi/minecraftinterface/legacy/BetaMinecraftInterface.java
```diff
@@ -39,8 +63,49 @@ private BetaMinecraftInterface(SymbolicClass biomeClass, SymbolicClass dimension
...
+
+    @Override
+    public RecognisedVersion getRecognisedVersion() {
+        return recognisedVersion;
+    }
+
+    private Class<?> makeInstantiableDimension(SymbolicClass dimensionBaseClass) {
+        // Since b1.6, the Dimension class is abstract, with OverworldDimension
+        // being an empty class inheriting from it. Our ClassTranslator cannot
+        // match "empty class that inherits from Overworld", so we just generate
```
burgerguy, line 98: What makes this true? Can't we just check for the amount of methods, fields, and constructors to see if they match what it should, and then see if it extends what we need?

burgerguy:
Treer and I have both separately tried to create a class to do this, but I noticed that this one has a really good solution for displaying oceans. I'll take a look at the source when I get the chance, because this looks like it could be really nice. However, moulins may not accept this until we're certain that ByteBuddy is absolutely required.

src/main/java/amidst/mojangapi/minecraftinterface/legacy/BetaMinecraftInterface.java
kahomayo: The main problem was the "extends what we need" part, as we currently only have a `getRealSuperClass` method available for matching. However, I now think that a `getSymbolicSuperClassName` method would be possible by giving `ClassTranslator.translate` a defined iteration order and supplying the already matched classes `result` to the `RealClassDetector`. This would mean that class detection will not be able to be parallelized in the future, but it doesn't look like this was planned.

I'll try to make the changes outlined above and push a commit in the coming days.

kahomayo:
These two commits remove the Byte Buddy dependency by adding some ClassTranslator functionality and implementing the ocean finding algorithm completely in Amidst.

burgerguy:
@moulins could you consider looking at this PR? It looks really good and polished, and it would be nice to finally close the old beta support branch.

moulins:
Oh sorry, I completely forgot this, haha!

Everything looks good, so I'll merge it once I have the time to quickly test this :)

moulins (approved):

src/main/java/amidst/mojangapi/minecraftinterface/legacy/BetaMinecraftInterface.java
```diff
@@ -0,0 +1,505 @@
...
+        return overworldLevelSourceClass.callConstructor(BetaSymbolicNames.CONSTRUCTOR_OVERWORLD_LEVEL_SOURCE, world.getObject(), worldOptions.getWorldSeed().getLong());
+    }
+
+    /** Provides both biomes and oceans, at a configurable precision. */
+    private class OceanProvidingWorldAccessor implements MinecraftInterface.WorldAccessor {
+        private final SymbolicObject biomeGenerator;
+        private final SymbolicObject overworldLevelSource;
+        private final boolean generateBlocks;
+        // instance variable to avoid allocations
+        private final byte[] blocks = new byte[32768];
```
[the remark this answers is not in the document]
kahomayo: Thank you for reviewing this, but you might have been reviewing old code (this diff was outdated). The same issue was also present in the latest version and has been fixed in "Make BetaWorldAccessor thread-safe".

moulins:
Thanks for your work @kahomayo!

